PR #319, opened 2014-12-09
Fixed #312: added support for group and host properties, backwards compatible

### LINKED ISSUE #312: Improve group and host deployment support
Recent changes allow adding @group on appservers and databases. It also creates groups if they don't exist yet. Unfortunately it doesn't assign hosts to those new groups, so databases assigned to those new groups end up without forests.

I recommend improving config by wrapping the relevant appservers inside gr:groups/gr:group with a gr:group-name that shows in which group those should be added. This provides space for setting group properties as well. Additionally a ho:hosts/ho:host structure needs to be added that allows setting host properties which also includes assigning a host to a group.

It is possible to rewrite configs on the fly to support backwards compatibility, and put appservers and database forest in Default group (or the one of the current host/app-server) if not explicitly specified.

grtjn:


grtjn:
Is the PR agreeable like this?

dmcassel:
good by me. @paxtonhare ?

PR #331, opened 2015-01-12
Adding roxy property to ignore certain files or folders

miguelrgonzalez:
Specially when working with node and bower there are certain folder we don't want to be deployed into the server.

### DIFF
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -204,3 +204,4 @@ mlcp-vmargs=-Xmx512m
 # Temporary fix to support MarkLogic 8 EA2
 #
 evaler-port=7998
+
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1130,12 +1130,14 @@ private
     load_html_as_xml = @properties['ml.load-html-as-xml']
     load_js_as_binary = @properties['ml.load-js-as-binary']
     load_css_as_binary = @properties['ml.load-css-as-binary']
+    folders_to_ignore = @properties['ml.ignore-folders']
 
     modules_databases.each do |dest_db|
       ignore_us = []
       ignore_us << "^#{test_dir}.*$" unless test_dir.blank? || deploy_tests?(dest_db)
       ignore_us << "^#{app_config_file}$"
       ignore_us << "^#{test_config_file}$"
+      ignore_us << "^#{folders_to_ignore}$" unless folders_to_ignore.blank?
 
       src_permissions = permissions(@properties['ml.app-role'], Roxy::ContentCapability::ER)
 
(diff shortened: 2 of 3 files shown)


grtjn:
Have you considered keeping those outside src/?


Sent from Samsung Mobile

miguelrgonzalez:
Yes, I did.
But I prefer to have source code inside the src folder.
Plus having bower.js, gulp.js there also makes sense.

deploy/default.properties
```diff
@@ -204,3 +204,6 @@ mlcp-vmargs=-Xmx512m
 # Temporary fix to support MarkLogic 8 EA2
 #
 evaler-port=7998
+
+# folde to ignore when deploying into modules database
+ignore-folders=.*/src/(node_modules|.tmp)/.*
```
paxtonhare: Move this into build.sample.properties as a comment explaining how to use it. I don't think this should be the default for everyone.

deploy/lib/server_config.rb
```diff
@@ -1130,12 +1130,14 @@ def deploy_src
...
     load_js_as_binary = @properties['ml.load-js-as-binary']
     load_css_as_binary = @properties['ml.load-css-as-binary']
+    folders_to_ignore = @properties['ml.ignore-folders']
 
     modules_databases.each do |dest_db|
       ignore_us = []
       ignore_us << "^#{test_dir}.*$" unless test_dir.blank? || deploy_tests?(dest_db)
       ignore_us << "^#{app_config_file}$"
       ignore_us << "^#{test_config_file}$"
+	  ignore_us << "^#{folders_to_ignore}$"
```
paxtonhare: This should only append if folders_to_ignore is not blank.

paxtonhare: Also, check your usage of tabs versus spaces for consistency.

paxtonhare:
@miguelrgonzalez I like this and think it's useful. I added some inline comments that I'd like to see fixed before we merge it.

miguelrgonzalez:
Hi Paxton,
will do the requested changes and fix my commit.

paxtonhare:
Awesome. Thanks for submitting it.

On Mon, Jan 12, 2015 at 10:08 AM, Miguel Rodríguez <notifications@github.com

dmcassel:
+1 on this, once Paxton's comments are addressed

grtjn:
One sec before merging. I think it would be better to push this into the dev branch.

@miguelrgonzalez would it be a lot of trouble to ask you to redo your changes based on the dev branch, and do a PR against that one?

dmcassel:
Good catch, @grtjn

miguelrgonzalez:
No problem, it's just a couple of lines really.

grtjn:
Replaced by PR #337..

Issue #335, opened 2015-01-14
Raise exception to OS level for integratability


bobstarbird:
Add "raise e" to ml.rb to take advantage of Ruby setting exit status so ML commands can be embedded and exit status tested externally

rescue Net::HTTPServerException => e
  case e.response
  when Net::HTTPUnauthorized then
    @logger.error "Invalid login credentials for #{@properties["environment"]} environment!!"
    raise e
  else
    @logger.error e
    @logger.error e.response.body
    raise e
  end
rescue Net::HTTPFatalError => e
  @logger.error e
  @logger.error e.response.body
    raise e
rescue DanglingVarsException => e
  @logger.error "WARNING: The following configuration variables could not be validated:"
  e.vars.each do |k,v|
    @logger.error "#{k}=#{v}"
  end
    raise e
rescue HelpException => e
  Help.doHelp(@logger, e.command, e.message)
    raise e
rescue ExitException => e
  @logger.error e
    raise e
rescue Exception => e
  @logger.error e
  @logger.error e.backtrace
    raise e
end

dmcassel:
Sounds like a good idea. @paxtonhare, if you agree I'll do a PR.

grtjn:
+1

paxtonhare:
Go for it

grtjn:
Fixed in dev

PR #338, opened 2015-01-15
#335 adding 'raise e' so that exit statuses will be returned

dmcassel:


dmcassel:
@paxtonhare, since I submitted this one, can you review and accept?

paxtonhare:
I'll go on a Roxy pull spree later this morning.

PR #340, opened 2015-01-16
Fixed #336: error-format property

### LINKED ISSUE #336: ML8 doesn't support REST property error-format
Using a new Roxy REST project against MarkLogic 8 will give this error at deploy rest:

```
ERROR: 400 "Bad Request"
ERROR: {"errorResponse":{"statusCode":400, "status":"Bad Request", "messageCode":"RESTAPI-UNSUPPORTEDPROP", "message":"RESTAPI-UNSUPPORTEDPROP: (err:FOER0000) Unsupported property:  : error-format"}}

```
Workaround: get rid of that line in rest-api/config/properties.xml. We should get rid of that line in the sample version of that file..

grtjn:


dmcassel:
Does MarkLogic <= 7 respond well if this property is missing? Would it better to comment it out for now with a note that it's only applicable for MarkLogic versions 6 and 7?

grtjn:
ML7 responds well, so I guess 6 also. I will put it in a comment..

grtjn:
Something like this?

dmcassel:
perfect

PR #351, opened 2015-01-20
Fixed #322: preserving retired forest at bootstrap

### LINKED ISSUE #322: Bootstrap looses retire status of forest
Run bootstrap with content-forests-per-hosts=3, then retire one of three, rerun bootstrap. Retire flag is effectively removed, where I think it should have been preserved..

This is only the case for forests automatically added by Roxy. Marking forest as Read-only is preserved..

grtjn:


dmcassel:
I think this looks good, but I'm not sure why it was written the other way before. @paxtonhare?

paxtonhare:
@dmcassel Not sure why it was done the other way. I'm going with this one though.

Issue #353, opened 2015-01-25
Capture configs is unable to capture users and roles when specified by name


alencruz:
The setup:get-configuration function in setup.xqy used only ids to return users and roles. This prevented the capture configurations option from capturing users and roles specified by name in the current project configuration or as a parameter. This was a documented “TODO” in the code in order to improve the capture functionality. 
I have added this capability and will be submitting a pull request momentarily. The setup:get-configuration function will now take user/role names or ids as a parameter and return the requested configurations.

grtjn:
Fixed in dev

PR #354, opened 2015-01-25
Fixing Issue #353 Capture config users&roles by name

alencruz:
The setup:get-configuration function in setup.xqy used only ids to
return users and roles. This prevented the capture configurations
option from capturing users and roles specified by name in the
current project configuration or as a parameter. This was a
documented “TODO” in the code in order to improve the capture
functionality. The setup:get-configuration function will now take
user/role names or ids as a parameter and return the requested
configurations. Also added default users and roles from current
project to capture_environment_config in server_config.rb. Lastly
removed word “full” from logging since it could be a full or partial
configuration that is being captured.

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -4562,7 +4562,7 @@ declare function setup:get-privilege-by-name($name as xs:string) as element(sec:
 declare function setup:get-users-by-name($names as xs:string*) as element(sec:users)? {
   let $ids :=
     for $name in $names
-      return setup:get-user-id($name)
+      return xdmp:user($name)
   return setup:get-users($ids)
 };
 
@@ -4633,7 +4633,7 @@ declare function setup:get-user-id($user-name as xs:string) as xs:unsignedLong? 
 declare function setup:get-roles-by-name($roles as xs:string*) as element(sec:roles)? {
   let $ids :=
     for $role in $roles
-      return setup:get-role-id($role)
+      return xdmp:role($role)
   return setup:get-roles($ids)
 };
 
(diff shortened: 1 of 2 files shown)


dmcassel:
Looks good (haven't tested). A future enhancement would be to pull in the names from the merged configuration rather than the properties, as there are sometimes users or roles put directly into ml-config.xml.

alencruz:
Simplified implementation based on Geert's comments, thanks Geert!

deploy/lib/xquery/setup.xqy
```diff
@@ -4544,6 +4559,13 @@ declare function setup:get-privilege-by-name($name as xs:string) as element(sec:
     </options>)
 };
 
+declare function setup:get-users-by-name($names as xs:string*) as element(sec:users)? {
+  let $ids :=
+    for $name in $names
+      return xdmp:user($name)
```
grtjn: I am afraid my advise has had a negative side effect. You former approach would probably have silently ignored users and roles that didn't exist, whereas this will throw a large stacktrace. I lean towards ignoring it silently, you will notice soon enough it didn't capture such users/roles once you look inside the captured ml-config..

deploy/lib/xquery/setup.xqy
```diff
@@ -4282,18 +4282,33 @@ declare function setup:get-configuration(
...
-  <configuration>
-    {setup:get-app-servers($app-servers)}
-    {setup:get-forests($forests)}
-    {setup:get-databases($databases)}
-    {setup:get-users($user-ids)}
-    {setup:get-roles($role-ids)}
-    {setup:get-mimetypes($mimetypes)}
-  </configuration>
+  let $user-configuration :=
+      typeswitch($users[1])
```
grtjn: The typeswitch looks nice, but doesn't work like this. I think you should use an old fashion if with a castable as test, and then actually typecast the args for setup:get-users. I would also recommend checking and casting to xs:unsignedLong, as that is what the get-users function expects..

dmcassel:
Is this PR still in progress?

alencruz:
My apologies, I had not seen the last comments from Geert, my first foray into GitHub, I will look over all comments within the next couple of days for both of my pull requests and report back findings/commit fixes as needed.

deploy/lib/xquery/setup.xqy
alencruz: Yup, I see that now, I had tested by injecting strings vs. integers from the Ruby capture_environment_config function directly which worked ok, however it does look like if the values are passed in as arguments they are quoted to strings. I will fix this shortly. Thanks!

deploy/lib/xquery/setup.xqy
alencruz: I agree, I will make change to absorb when users and roles don't exists. Thanks!

alencruz:
Committed changes based on feedback.

grtjn:
Capture now runs without errors..

Issue #298, opened 2014-11-03
capture of modules-db does not work anymore


peetkes:
I tried to capture the content of a modules-db but that one failed.
./ml test capture --modules-db=sarepta-modules
Retrieving source and REST config from sarepta-modules...
ERROR: No such file or directory @ rb_sysopen - /var/folders/hj/45wrdfss2ll8jhqyy0c2r32c24hql1/T/d20141103-65428-1kxoe0j/src/app/config/config.xqy
ERROR: ["/Users/pkester/Development/StandardNorway/capture/deploy/lib/server_config.rb:1024:in `initialize'", "/Users/pkester/Development/StandardNorway/capture/deploy/lib/server_config.rb:1024:in `open'", "/Users/pkester/Development/StandardNorway/capture/deploy/lib/server_config.rb:1024:in `block in save_files_to_fs'", "/Users/pkester/Development/StandardNorway/capture/deploy/lib/server_config.rb:1017:in `each'", "/Users/pkester/Development/StandardNorway/capture/deploy/lib/server_config.rb:1017:in `save_files_to_fs'", "/Users/pkester/Development/StandardNorway/capture/deploy/lib/server_config.rb:946:in `capture'", "deploy/lib/ml.rb:140:in `<main>'"]

The mentioned folder exists but is completely empty.
Talked to Geert about this and apparently the folders are not created before writing the file.
This is probably caused by this function in server_config.rb:
  def save_files_to_fs(target_db, target_dir)
    # Get the list of URIs. We get them in order because Ruby's Dir.mkdir
    # command doesn't have a -p option (create parent).

I changed this (line 1024):
          File.open("#{target_dir}#{uri}", 'wb') { |file| file.write(r.body) }
into this (thanks to Geert!)
          path = "#{target_dir}#{uri}"
          parentdir = File.dirname "#{target_dir}#{uri}"
          FileUtils.mkdir_p(parentdir) unless File.exists?(parentdir)
          File.open("#{path}", 'wb') { |file| file.write(r.body) }
and then it worked again.
Same should apply to line 1006

grtjn:
Fixed in dev

Issue #223, opened 2014-04-29
Allow capturing modules for any app-type


grtjn:
You sometimes want to capture an existing Roxy app from a production server, for example to check whether what you locally have is identical. This requires capturing the modules of any type.

alencruz:
I have added the ability to capture modules for any application type to ML7 and taken the protection off which will allow to capture modules for any application type in ML5,6, and 7. For ML7 if it’s a REST application, REST will still be used to capture since it’s faster and more comprehensive than the query based method. Pull request to follow.

grtjn:
Fixed in dev

Issue #163, opened 2014-01-06
ml capture doesn't get binary files correctly


dmcassel:
When "ml capture" against a ML7 instance, results are expected to be JSON. That works for text content, but not for binary. See the save_files_to_fs function in deploy/lib/server_config.rb.

PR #356, opened 2015-01-27
Fixing Issue #223 Capture modules for any app type

alencruz:
Currently Capture was only supporting capturing modules for
REST applications. I have added the ability to capture modules
for any application type to ML7 and taken the protection off
which will allow to capture modules for any application type in
ML5,6, and 7. For ML7 if it’s a REST application, REST will still be
used to perform the capture since it’s faster than the query
based method. The query based method used for ML5,6, and
ML7 (non-REST) only grabs text based files (xqy, html,…) it does
not correctly capture other types of files in the modules database
such as images since it’s doing an fn:doc on the URI to get the
content.

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -952,23 +952,29 @@ In order to proceed please type: #{expected_response}
         logger.info "Retrieving source from #{target_db}..."
       end
 
-      # send the target db, the destination directory, and port to know if REST for ML7+
-      save_files_to_fs(target_db, "#{tmp_dir}/src", port)
+      # send the target db, and the destination directory
+      save_files_to_fs(target_db, "#{tmp_dir}/src")
+		
+	  # check if this is a REST project to capture REST configuration
+	  if (port != nil)	
+	  
+		  # make sure that REST	options directory exists
+		  if Dir.exists? @properties['ml.rest-options.dir'] 
+		  
+			# set up the options
+			FileUtils.cp_r(
+			  "#{tmp_dir}/src/#{@properties['ml.group']}/" + target_db.sub("-modules", "") + "/rest-api/.",
+			  @properties['ml.rest-options.dir']
+			)
+			FileUtils.rm_rf("#{tmp_dir}/src/#{@properties['ml.group']}/")
 
-      if (port != nil)
-        # set up the options
-        FileUtils.cp_r(
-          "#{tmp_dir}/src/#{@properties['ml.group']}/" + target_db.sub("-modules", "") + "/rest-api/.",
-          @properties['ml.rest-options.dir']
-        )
-        FileUtils.rm_rf("#{tmp_dir}/src/#{@properties['ml.group']}/")
-      end
-
-      # Make sure REST properties are in accurate format, so you can directly deploy them again..
-      if (port != nil)
-        r = go("http://#{@hostname}:#{port}/v1/config/properties", "get")
-        r.body = parse_json(r.body)
-        File.open("#{@properties['ml.rest-options.dir']}/properties.xml", 'wb') { |file| file.write(r.body) }
+			# Make sure REST properties are in accurate format, so you can directly deploy them again..
+			r = go("http://#{@hostname}:#{port}/v1/config/properties", "get")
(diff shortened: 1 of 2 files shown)


grtjn:
The changes look good. I'll try to test asap..

deploy/lib/server_config.rb
```diff
@@ -947,16 +945,24 @@ def capture
...
-        "#{tmp_dir}/src/#{@properties['ml.group']}/" + target_db.sub("-modules", "") + "/rest-api/.",
-        @properties['ml.rest-options.dir']
-      )
-      FileUtils.rm_rf("#{tmp_dir}/src/#{@properties['ml.group']}/")
+      # send the target db, the destination directory, and port to know if REST for ML7+
+      save_files_to_fs(target_db, "#{tmp_dir}/src", port)
+
+      if (port != nil)
+        # set up the options
+        FileUtils.cp_r(
```
grtjn: I'm hitting issue #298. If you start with mvc app-type, the rest-api dir doesn't exist yet. There is code mentioned in the ticket that you could include here.

deploy/lib/server_config.rb
```diff
@@ -947,16 +945,24 @@ def capture
     if target_db != nil
       tmp_dir = Dir.mktmpdir
       logger.debug "using temp dir " + tmp_dir
-      logger.info "Retrieving source and REST config from #{target_db}..."
 
-      save_files_to_fs(target_db, "#{tmp_dir}/src")
+      if (port != nil)
```
grtjn: Nice way to solve capturing non-REST modules..

deploy/lib/server_config.rb
```diff
@@ -1026,8 +1032,22 @@ def save_files_to_fs(target_db, target_dir)
           # create the directory so that it will exist when we try to save files
           Dir.mkdir("#{target_dir}" + uri)
         else
-          r = go("#{@protocol}://#{@hostname}:#{@bootstrap_port}/qconsole/endpoints/view.xqy?dbid=#{db_id}&uri=#{uri}", "get")
-          File.open("#{target_dir}#{uri}", 'wb') { |file| file.write(r.body) }
+          # check type, if REST use that endpoint, otherwise use a query
+          if (port != nil)
```
grtjn: I think this is unnecessary. The port value is not used, and the downloading of files depends on QConsole, which works for any app-type..

deploy/lib/server_config.rb
```diff
@@ -1026,8 +1032,22 @@ def save_files_to_fs(target_db, target_dir)
...
+            r = go("#{@protocol}://#{@hostname}:#{@bootstrap_port}/qconsole/endpoints/view.xqy?dbid=#{db_id}&uri=#{uri}", "get")
+            file_content = r.body
+          else
+            # Note : There are limitations to doing it this way for non REST DBs; only text files are captured (xqy, html, etc..) 
+            # other types such as images or java scripts will not be grabbed. Also this method is slower than REST or XCC.
```
grtjn: I think that is why @dmcassel moved away from this method in #163. And using QConsole approach of above also bypasses this problem I think. I am pretty sure the capture was capturing images of app-builder apps correctly..

dmcassel: The problem in #163 was that the REST API (at least at the time) was returning a JSON description of the binary files, instead of the file itself.

deploy/lib/server_config.rb
alencruz: I like throwing the error idea, that way the user does not end up with something unexpected. I have implemented this and will be committing momentarily. I will note that it also fixes issue #298 since "capture" will no longer try to save REST configuration if the folder does not exists in the project.

deploy/lib/server_config.rb
alencruz: Agree, I have removed this now.

alencruz:
Committed changes based on feedback.

grtjn:
The indentation of the help looks a little garbled, but otherwise ok. I'll merge, we can touch help later..

grtjn:
Sorry, I think I meant to merge, but clicked Close-n-comment instead..

PR #382, opened 2015-02-20
Fixed #362: support for app-type bare

### LINKED ISSUE #362: Support a new type of application for those which are neither REST nor MVC
MarkLogic is way too versatile to see its applications fall follow one of these two design patterns and while it is [possible to package and deploy generic applications with Roxy](http://stackoverflow.com/questions/28278234/can-roxy-be-used-to-deploy-generic-marklogic-applications/28279288) this would be more straightforward with a new type of application (which could be called "raw", "generic", ...) which would behave like MVC but generate an empty `src` sub-directory.

grtjn:


dmcassel:
Looks good to me.

PR #385, opened 2015-02-20
Fixed #377: Added support for deploy rest of .sjs, and some cleanup

### LINKED ISSUE #377: Deploy rest ignores .sjs extensions

grtjn:
Special thanks to @steliann for providing working code. Only rearranged a bit, and did some cleanup..

### DIFF
--- a/deploy/lib/ml_rest.rb
+++ b/deploy/lib/ml_rest.rb
@@ -18,7 +18,7 @@ module Roxy
     def get_files(path, data = [])
       @logger.debug "getting files for #{path}"
       if (File.directory?(path))
-        Dir.glob("#{path}/*.{xqy,xslt,xsl}") do |entry|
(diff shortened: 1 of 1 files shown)


deploy/lib/ml_rest.rb
```diff
@@ -122,37 +173,55 @@ def install_extensions(path)
...
+          }
+          
+          transformName = file_name
           params = []
-
-
-          # TODO: I'm assuming there's a way to consolidate the following if/elsif but I'm a Ruby newbie and I'm being conservative
-          if (file_type == 'xquery')
+          
+          if (is_sjs)
```
grtjn: This might need a second look. Didn't have a .sjs transform at hand, not sure it works properly..

paxtonhare: It's working. Just tested it.

PR #386, opened 2015-02-23
Fixed #365: added howto help based on wiki pages

### LINKED ISSUE #365: Add howto command to show Roxy wiki pages
Something like: `./ml howto` or `./ml howto cpf`. The latter would search for wiki page titles containing that string. Not as good as searching the page contents, but github has no search for wiki pages.

I experimented a bit with nokogiri and the built-in _pages wiki page:

```ruby
#!/usr/bin/ruby

begin
  require 'open-uri'
  require 'rubygems'
  require 'nokogiri'

  def howto
    search = ARGV.first
    doc = Nokogiri::HTML(open('https://github.com/marklogic/roxy/wiki/_pages'))

    pages = doc.css('.content').select do |page|
      search == nil or page.text.downcase().include? search
    end

    selected = 1

    if pages.length > 1
      count = 0
      pages.each do |page|
        count = count + 1
        puts "#{count} - #{page.text}\n\thttps://github.com/#{page.xpath('a/@href').text}"
      end
  
      print "Select a page: "
      selected = STDIN.gets.chomp().to_i
      if selected == 0
        exit
      end
      
      if selected > pages.length
        selected = pages.length
      end
    end

    count = 0
    pages.each do |page|
      count = count + 1
      if count == selected
    
        puts "\n#{page.text}\n\thttps://github.com/#{page.xpath('a/@href').text}"
    
        uri = "https://github.com/#{page.xpath('a/@href').text}"
        doc = Nokogiri::HTML(open(uri))

        puts doc.css('.markdown-body').text.gsub(/\n\n\n+/, "\n\n")
    
      end
    end
  end

  howto

rescue LoadError => e
  puts "Missing library: #{e}"
end
```

This is a self-contained script that works. Just need a nice way to integrate into Roxy. Issue is that nokogiri could be missing, and not sure how to gracefully decay when it is embedded in server_config.rb or alike..

@paxtonhare maybe you have an idea how to encapsulate this such that it doesn't break whole ml?

grtjn:


dmcassel:
This PR adds requires for 'open-uri' and 'nokogiri'. We've always tried to minimize the number of gems required for Roxy to work. In this case, if the gems aren't available, the require is protected; it's just the new howto feature that won't work. @paxtonhare, thoughts?

grtjn:
Correct. I saw no easy way to do this without nokogiri. open-uri seems to be commonly available, I explicitly tested on demo servers with old Ruby (1.8.7)..

grtjn:
Yes correct. I had the include in the function itself before, but then noticed there were optional includes at the top already, and thought that might be cleaner. I could change around..

dmcassel:
I think I'd prefer the include being in the function. If it's at the top, then it will look like it's generally available and end up becoming a requirement. Our approach since the beginning has been to minimize the dependencies, so I think it's best to contain that one.

grtjn:
Sure no prob. I'll update the PR tomorrow..

grtjn:
Moved the optional requires to inside the howto function. Good to go?

PR #393, opened 2015-02-23
Fixed #376: --no-prompt param suppresses user queries

### LINKED ISSUE #376: Suppressing prompts
Running Roxy from for instance Slush sometimes stops without user feedback because of prompts. Slush doesn't pass stdin/stdout through, and user is unable to answer questions from Roxy, he/she isn't even aware of.

A command-line option to override asking for input could prevent such 'dead-locks'. Maybe `ml -f` or `ml --headless`?

grtjn:


grtjn:
I have rebased, this can now be merged automatically again..

dmcassel:
One prompt I don't see addressed is prompting for password, when the properties files don't have it. I guess we're assuming that won't happen in the slush case, but a parameter called "no-prompt" ought to get that one, too. Agree?

grtjn:
I was focussing on the slush case indeed. You also have creation of indexes that is doing user prompts. Instead of suppressing all of those, I left those as they were since you explicitly call those I thought.

It would make sense to do some assertion though, and exit with cannot prompt for password, or cannot prompt for input..

dmcassel:
For the message, I would say "--no-prompt parameter prevents prompting for password" (or input) or something along those lines -- indicating that's why it can't. Sounds good.

grtjn:
Better this way?

Note: Indentation changes make it appear more lines have been edited than have really been touched..

dmcassel:
Looks good.

PR #406, opened 2015-03-19
Fixed #405: support for field-value-searches on fields

### LINKED ISSUE #405: Make sure field-value-searches works at field level
field-value-searches should work at the database level and the field level, but we have a report of it not working at the field level when added to ml-config.xml. This ought to work:

    <field>
      <field-name>verladerNummer</field-name> 
      <field-path> 
        <path>PcsRef[CRlTyCd eq "VL"]/PcsRef</path> 
        <weight>1</weight>
      </field-path>
      <word-lexicons/>
      <tokenizer-overrides/>
      <field-value-searches>true</field-value-searches>
      <trailing-wildcard-searches>true</trailing-wildcard-searches>
      <included-elements/>
      <excluded-elements/>
    </field>

grtjn:


dmcassel:
Looks good Geert, thanks!

PR #424, opened 2015-04-30
Fixed #359: added separate mlcp user/pass properties

### LINKED ISSUE #359: Make it easy to run MLCP with non-admin user
If I want to load data via mlcp as a non-admin user I can't unless I hand jam server_config.rb or run mlcp directly. I threw a quick fix into my current project, but this should be slightly more extensible.

server_config.rb line 847ish....

```ruby
username = @properties['ml.default-user']
password = @properties['ml.appuser-password']
if ARGV.length > 0
  password_prompt
  connection_string = %Q{ -username #{username} -password #{password} -host #{@properties['ml.server']} -port #{@properties['ml.xcc-port']}}
```

Ideally I'd have a mlcp.user= and mlcp.password= so I can overwrite them with anything.

default them to:
```
mlcp.user=${user}
mlcp.password=${password}
```
then let users do whatever in their properties files.

grtjn:
Leveraging @ml_user and @ml_pass to make full use of password_prompt..

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -989,9 +989,11 @@ In order to proceed please type: #{expected_response}
       end
     end
 
+    @ml_username = @properties['ml.mlcp-user']
+    @ml_password = @properties['ml.mlcp-password']
     if ARGV.length > 0
       password_prompt
-      connection_string = %Q{ -username #{@properties['ml.user']} -password #{@ml_password} -host #{@properties['ml.server']} -port #{@properties['ml.xcc-port']}}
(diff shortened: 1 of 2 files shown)


deploy/lib/server_config.rb
```diff
@@ -989,9 +989,11 @@ def mlcp
       end
     end
 
+    @ml_username = @properties['ml.mlcp-user']
+    @ml_password = @properties['ml.mlcp-password']
```
paxtonhare: Just as an extra precaution, perhaps these should be:

@ml_username = @properties['ml.mlcp-user'] || @properties['ml.user']
@ml_password = @properties['ml.mlcp-password'] || @ml_password

Not everyone knows about or uses the ./ml upgrade feature and thus might not get those properties added to default.properties.

grtjn:
Done!

Issue #427, opened 2015-05-01
Help folks stuck on ruby 1.8.7


paxtonhare:
A lot of Roxy users are stuck on ruby 1.8.7 for whatever reason. Try to make their lives a bit easier.

Here's the first one:
ERROR: undefined method `body=' for #<Net::HTTPOK 200 OK readbody=true>
ERROR: ./deploy/lib/server_config.rb:425:in `bootstrap'deploy/lib/ml.rb:144:in `send'deploy/lib/ml.rb:144

I have a patch. Just need to do a pull request.

grtjn:
+10! There might be other tickets talking about things like this..

paxtonhare:
I'm closing this until other issues appear

PR #428, opened 2015-05-01
issue #427 - patching Ruby code with missing method

paxtonhare:


grtjn:
Cool, monkey patching rules!

Issue #431, opened 2015-05-03
triggers-db = modules-db with version 1.7.2


rlouapre:
In Roxy 1.7.2 I am not sure how to setup ml-config.xml and properties to use triggers-db = modules-db

See my comment https://github.com/grtjn/roxy/commit/902810059f24166fdca82c9bc97bdf9c540fb9b1

This suggestion https://github.com/marklogic/roxy/issues/361#issuecomment-72929826 does not seem to work.

paxtonhare:
I left a comment in the other place but this is a legit find. Thanks @rlouapre .

grtjn:
I'll take a close look asap..

grtjn:
Fixed in dev

PR #432, opened 2015-05-04
issue #431- fixing triggers-db = modules-db

paxtonhare:
Fixes issue #431


grtjn:
Hmm, odd. Looks like you try to suppress db creation, but not the forests? I'd expect you would not need either. Triggers get stored in mod db and thus mod forests.

But maybe i should take a close look from laptop instead of just staring at code from a tiny phone screen..

PR #440, opened 2015-05-15
Fix #437

### LINKED ISSUE #437: Export properties for a given environment
I am looking for a way to export the list of propertie for a given environment and I was planning to use ```ml info``` for that. 
The idea is to reuse the parameter defined in Roxy for other tools (gulp, mocha...).
The current format used by ```ml info``` can be easily formatted but we need to be careful with providing more feedback https://github.com/marklogic/roxy/issues/414#issuecomment-100589864.
It would be event easier if Roxy could support standard format: ```xml```, ```csv``` or ```yml```.
Or is there a better way than using ```ml info```

@paxtonhare - how MarkLogic Sublime does it?

rlouapre:
Successfully tested with xmllint and jq
ml {env} info --format support xml and json

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -102,10 +102,27 @@ class ServerConfig < MLClient
   end
 
   def info
-    logger.info "IS_JAR: #{@@is_jar}"
-    logger.info "Properties:"
-    @properties.sort {|x,y| y <=> x}.each do |k, v|
-      logger.info k + ": " + v
+    format = find_arg(['--format'])
+    info = {}
+    info["isJar"] = @@is_jar
+    info["properties"] = @properties
+    if format == "json"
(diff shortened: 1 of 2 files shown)


deploy/lib/server_config.rb
```diff
@@ -102,10 +102,27 @@ def get_properties
   end
 
   def info
-    logger.info "IS_JAR: #{@@is_jar}"
-    logger.info "Properties:"
-    @properties.sort {|x,y| y <=> x}.each do |k, v|
-      logger.info k + ": " + v
+    format = find_arg(['--format'])
+    info = {}
```
grtjn: You can move this variable inside the if, otherwise clean and simple.

grtjn:
Would be nice to add it to Help.rb as well, for completeness.

grtjn:
Fixed #437

PR #453, opened 2015-06-01
fixed property name for getting rest modules db

ryan321:
Other places in the code the property name "ml.app-name-rest-modules' is used instead of ['ml.app-name']}-rest-modules. This isn't a problem if the properties resolve to the same value. But I have a situation where the rest-modules db is the same as the app modules db ( in build.properties I have: app-name-rest-modules=${app-name}-modules). So then an error is thrown on deployment that the rest db is not found because the assumption is that the rest db name will be {app-name}-rest-modules.

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1534,7 +1534,7 @@ private
       # Figure out where we need to deploy this stuff
       rest_modules_db = ''
       if @properties.has_key?('ml.rest-port') and @properties['ml.rest-port'] != ''
-        rest_modules_db = "#{@properties['ml.app-name']}-rest-modules"
+        rest_modules_db = "#{@properties['ml.app-name-rest-modules']}"
       else
         rest_modules_db = @properties['ml.modules-db']
       end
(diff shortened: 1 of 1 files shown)


deploy/lib/server_config.rb
```diff
@@ -1534,7 +1534,7 @@ def deploy_rest
       # Figure out where we need to deploy this stuff
       rest_modules_db = ''
       if @properties.has_key?('ml.rest-port') and @properties['ml.rest-port'] != ''
-        rest_modules_db = "#{@properties['ml.app-name']}-rest-modules"
+        rest_modules_db = "#{@properties['ml.app-name-rest-modules']}"
```
grtjn: You can simplify this to:

```
rest_modules_db = @properties['ml.app-name-rest-modules']
```

grtjn:
Looks good, and nice find that rewriter for rest-port! I have just one minor comment. Can you append to the PR? Just commit to your dev branch and push to github, it will show up here automatically..

grtjn:
Hmm, on second thought, There is no such property ml.app-name-rest-modules. I think Ryan mistakenly took @ml.app-name-rest-modules as one property name, but it is actually really @{ml.app-name}-rest-modules..

I think we need to revert or fix this, and come up with a dedicated property for rest modules, maybe rest-modules-db..

Issue #414, opened 2015-04-14
Allow users to set properties via OS environment variables


paxtonhare:
Provide a mechanism for letting users set properties to environment variables.

I created this as a workaround: https://github.com/marklogic/roxy/wiki/Using-Operating-System-Environment-Variables-in-properties

But I'd like to see it become an included feature for .properties. Perhaps something like:

```
user=env['app_username']
password=env['app_password']
```

maybe even (note the lack of quotes):

```
user=env[app_username]
password=env[app_password]
```

You could even allow other variables to get subbed-in:

```bash
export app_username=admin
export app_local_username=admin2
```

```
app-name=my-app
user=${app-name}-env[app_username]   # would become user=my-app-admin
password=env[app_${environment}_password] # would become user=admin2
```

grtjn:
How about something a little more dense like:

```
user=%{MYENVUSER}
```

rlouapre:
How about using the same syntax used for passing properties from command line?
Property value ```ml.password``` will be replaced by the environment variable ```ml.password```

paxtonhare:
I like the idea of always checking the environment variables of the same
name before using the value from the properties files.

pseudo code would be:

property[key] = env[key] || @properties[key]

The only downside I see to this approach is that you could drive yourself
mad if you accidentally set the property in the shell and forget you did
it. Then you keep changing the value in the .properties file and wonder why
it's not being picked up.

grtjn:
Sounds interesting. I like the idea of prefixing with ml. We could also extend ml local info to provide more feedback, like where value came from, env, env.prop, build or defaults..

rlouapre:
Could ```logger.debug``` provide detail on where the property value comes from?
Also to clarify property can be defined from environment variable, command line argument and property file.
Should the priority be the same listed above?
That would be a great improvement for CI integration.

grtjn:
debug versus info, nice idea. Yes, I think debug should mention name of properties file (local.properties/build.properties), or `ENV` or `cmd-line`. Command-line works strongest, then ENV, then properties in existing order.

Not sure what you mean by priority?

rlouapre:
Assuming ```ml.property1``` is set as environment variable, command line argument and file properties which has the highest priority.

grtjn:
[...]

PR #478, opened 2015-06-30
Fixed #467: allowed multiple prop replace syntax

### LINKED ISSUE #467: Allow same prop replace syntax in props and ml-config
In properties files you use ${app-name}, in ml-config you use @ml.app-name. The difference is confusing to my opinion. And not adding { } in ml-config makes people believe @ml.app-name-rest-modules is a full property name, where it actually is just a partial one (@ml.app-name + '-rest-modules').

I vote for adding different syntax to ml-config as well, preferably ${propname} like in properties themselves, alternatively @{ml.propname}. Thoughts?

grtjn:
Fixes #467

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -116,7 +116,7 @@ class ServerConfig < MLClient
       logger.info "<info>"
       logger.info "\s\s<isJar>#{@@is_jar}</isJar>"
       logger.info "\s\s<properties>"
-      @properties.sort {|x,y| y <=> x}.each do |k, v|
+      @properties.sort {|x,y| x <=> y}.each do |k, v|
         logger.info "\s\s\s\s<property name=\"#{k}\">#{v}</property>"
       end
       logger.info "\s\s</properties>"
@@ -124,7 +124,7 @@ class ServerConfig < MLClient
     else
       logger.info "IS_JAR: #{@@is_jar}"
       logger.info "Properties:"
-      @properties.sort {|x,y| y <=> x}.each do |k, v|
+      @properties.sort {|x,y| x <=> y}.each do |k, v|
         logger.info k + ": " + v
       end
     end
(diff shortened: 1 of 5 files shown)


grtjn:
I didn't replace all placeholders to distinguish between real properties, and inserts of pieces of config. Only exception is ssl cert stuff. I replaced that in a different PR by an insert as well.

grtjn:
Note to self: consider including support for #414..

grtjn:
I actually did pile up #414, since it was just a one-line change..

deploy/lib/server_config.rb
```diff
@@ -112,7 +112,7 @@ def info
       logger.info "<info>"
       logger.info "\s\s<isJar>#{@@is_jar}</isJar>"
       logger.info "\s\s<properties>"
-      @properties.sort {|x,y| y <=> x}.each do |k, v|
+      @properties.sort {|x,y| x <=> y}.each do |k, v|
```
dmcassel: If I remember right, the reason we put this sort in place is to make sure we replace the longest properties first. That way, if we see "@ml.app-name-foo-bar" and we have properties "app-name" and "app-name-foo", we want to use "app-name-foo" for the substitution. 

What's the reason for changing the sort order?

deploy/lib/server_config.rb
```diff
@@ -2172,12 +2164,32 @@ def build_config(config_file)
...
+    # escape unresolved braces, they have special meaning in XQuery
+    config.gsub!("{", "{{")
+    config.gsub!("}", "}}")
 
     config
   end
+  
+  def replace_properties(contents, name)
+    @properties.sort {|x,y| y <=> x}.each do |k, v|
+      # new property syntax
```
dmcassel: Adding a comment showing the new property syntax will be helpful down the road.

deploy/lib/server_config.rb
```diff
@@ -2172,12 +2164,32 @@ def build_config(config_file)
...
+      n = k.sub("ml.", "")
+      contents.gsub!("@{#{n}}", v)
+      contents.gsub!("${#{n}}", v)
+      
+      # backwards compat
```
dmcassel: Adding a comment showing the old property syntax will be helpful down the road.

dmcassel:
Please rebase

deploy/lib/server_config.rb
grtjn: You are correct. Descending order is necessary for replacement, but these lines concern display/output only. Ascending order reads much easier..

I'll rebase and update the comments..

grtjn:
Better like this?

PR #479, opened 2015-07-06
Fixes #463 and #462: rewrote use of xdmp:server to include group

### LINKED ISSUE #462: Groups in ml-config are not preserved
groups in ml-config that have a non-default group name which is not used elsewhere in ml-config get lost in rewrite-config. It should include $import-configs/gr:groups/gr:group/gr:group-name in determining all unique group names..

### LINKED ISSUE #463: Cannot add servers with identical name in multiple groups
MarkLogic supports having app-servers with identical names in different groups, like Admin and Manage. Now that we have group support in Roxy, creating for instance xdbc servers in every group makes sense. But Roxy fails if you create app-servers with same name in different groups, because it looks up server id using xdmp:server which is expected (by Roxy) to return only 1 id. We need to go through xdmp:group-servers to figure out which is the correct one..

grtjn:
Fixes #462, and #463

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -407,6 +407,11 @@ declare function setup:suppress-comments($nodes) {
 (: for backwards-compatibility :)
 declare function setup:rewrite-config($import-configs as element(configuration)+) as element(configuration)
 {
+  setup:rewrite-config($import-configs, ())
+};
+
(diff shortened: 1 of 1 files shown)


grtjn:
Rebased to allow auto merge..

grtjn:
Pending since July, merging this. I'll run a full self-test tomorrow..

PR #485, opened 2015-07-07
Fixed #189: don't get port from browser for server-side localhost http-get

### LINKED ISSUE #189: Framework tests fail using ssh tunnels
It looks like if there's a mismatch between the port number your browser is hitting and the actual port the test appserver is running on, test:easy-url() returns the port your browser provides, rather than the correct port.

grtjn:
Fixes #189


grtjn:
Open since Jul, merging..

PR #488, opened 2015-07-08
Fixed #460: added merge and reindex command for content

### LINKED ISSUE #460: Add merge command for full merge
MarkLogic changed since 7 to keep deleted fragments for an hour before cleaning them on merge. You can force to get rid of all with:

```xqy
    xdmp:merge(
      <options xmlns="xdmp:merge">
        <merge-timestamp>{xdmp:request-timestamp()}</merge-timestamp>
      </options>)
```

Would be convenient to have that built in into Roxy as for instance `ml local merge content`..

grtjn:
Fixes #460

### DIFF
--- a/deploy/lib/Help.rb
+++ b/deploy/lib/Help.rb
@@ -32,8 +32,10 @@ class Help
         corb          Runs Corb against the given environment
         deploy        Loads modules, data, cpf configuration into the given environment
         load          Loads a file or folder into the given environment
+        merge         Merges a database on the given environment
         mlcp          Runs MLCP against the given environment
         recordloader  Runs RecordLoader against the given environment
+        reindex       Reindexes a database on the given environment
         settings      Lists all supported settings for a given environment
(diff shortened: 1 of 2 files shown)


deploy/lib/Help.rb
```diff
@@ -266,6 +268,32 @@ def self.deploy
...
+  def self.merge
+    <<-DOC.strip_heredoc
+      Usage: ml {env} merge WHAT [options]
+
+      General options:
+        -v, [--verbose]  # Verbose output
+
+      Please choose a WHAT below.
+
+        content     # Merges your content db in the given environment
```
dmcassel: Seems like supporting "modules" would be a trivial addition. Is it more complicated than I'm thinking? (This PR looks good, I'll accept as-is if you like.)

grtjn: Yes, likely. I could add modules, schemas, triggers like for deploy, but I didn't see much need for that. I can add anyhow..

Issue #499, opened 2015-07-18
Speedup Bootstraping of Configuration with Large Number of Indexes


alencruz:
In the configuration for our project we have a large number of indexes (around 700 indexes). We were experiencing a long time for the bootstrap to complete and were running into timeout errors. The timeout errors were fixed by increasing the Roxy http parameters but the underlying cause was still present. We made some tweaks to the setup.xqy file to perform index configurations (removal and creation) at once, rather than iteratively doing each index. This speed up our bootstrapping for our scenarios by around 90% and we were able to revert back to the default Roxy http parameters. I will be making a pull request with these changes to see if it’s something the project would like to utilize.

grtjn:
Looking forward to the PR!

grtjn:
Fixed in dev

Issue #492, opened 2015-07-08
setup.xqy fails to validate field range index config (most likely other range-indexes as well)


joemfb:
Since `setup.xqy` uses the XML snippets from `ml-config.xml` directly, without validating their contents, it's possible to end up with a schema-invalid `databases.xml` configuration. Once that happens the REST Management API throws errors for database updates.

Specifically, I had a database with a `range-field-index` element containing a `namespace-uri` element (from the dmlc-healthcare project). This bootstraps (and works) without issue via Roxy. When I try to configure a database using the REST Management API (using ml-gradle and OBI), the PUT to `/manage/v2/databases/obi-ui-content/properties` fails with a 400 error.

This may properly be a bug of the Admin API not validating it's parameters. I'm not sure what an appropriate resolution would be, but wanted to capture the problem here.

grtjn:
I thought we fixed most of these cases. Which version of Roxy?

joemfb:
This was on v1.7.1, but the implementation is the same on current master: https://github.com/marklogic/roxy/blob/master/deploy/lib/xquery/setup.xqy#L2447

grtjn:
Ah yes, from the looks we forgot to fix path-index, and field-index. Will look asap..

grtjn:
Still applies..

grtjn:
#499/#501 is related, but didn't fix this. Pushing away to allow release of 1.7.3 though..

RobertSzkutak:
Pushing off to the next release...

PR #501, opened 2015-07-20
Fixing Issue #499 Configuration with Large Number of Indexes

alencruz:
#499 With a large number of indexes the bootstrapping process was taking
a significant amount of time, beyond the default Roxy HTTP configuration.
After looking at setup.xqy it was found that some of the index removal
calls were iterating over the items instead of calling the removal function
with the entire set. Similarly, some of the index creation functions were
recursively adding the items instead of calling the admin function with the
entire set. After looking at docs for all supported ML versions it was found
that calling the functions with all items was supported.
Changes were made to remove and add indexes directly. This made a large
impact in the bootstrapping time for scenarios with a large number of
configuration items.
As a reference, in a test setup with around 1,200 configuration items the
bootstrap operation was reduced from around 20 minutes to under
2 minutes with the above being the only change.

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -1855,7 +1855,37 @@ declare function setup:add-range-element-indexes(
   $db-config as element(db:database)) as element(configuration)
 {
   admin:database-add-range-element-index(setup:remove-existing-range-element-indexes($admin-config, $database),
-    $database, $db-config/db:range-element-indexes/db:range-element-index)
+    $database,
+    setup:validated-range-element-indexes($db-config/db:range-element-indexes/db:range-element-index))
+};      
+        
+declare function setup:validated-range-element-indexes(     
+  $index-configs as element(db:range-element-index)*) as element(db:range-element-index)*     
+{
+  let $function := xdmp:function(xs:QName("admin:database-range-element-index"))      
+  let $validated-index-configs :=
+    for $index-config in $index-configs
+      return      
+        if (setup:at-least-version("6.0-1")) then     
+          xdmp:apply(     
+            $function,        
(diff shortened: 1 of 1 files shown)


grtjn:
Thanks for the PR! You took the effort eliminating recursion for all index types, which is helpful.

Unfortunately, you also eliminated a number of xdmp:eval's, and version checks that allow backwards-compatibility. I'd like to keep existing compatibility in as much as possible. There are still quite some consultants having to work with MarkLogic 6, and sometimes even 5 as well.

You also rewrote function calls to create index defs to copy them directly from the ml-config, but the ml-config is not validated, so there could be typos in there, and such. You have to be careful, as reported very recently by Joe: #492

alencruz:
Thank you for your feedback, I will work to address the items you highlighted and commit back to the PR to make these changes more backwards compatible and also look at the issue of configuration validation before sending to the Admin API.

alencruz:
Committed changes to PR to address where I had removed xdmp:evals, version checks, and validation in the original commit.

grtjn:
I'll run a few tests before merging, but looks very good at first glance. Doesn't solve #492 yet, but we can do that separately..

dmcassel:
@grtjn did you have a chance to test this?

grtjn:
not yet, will try to make some time on short term..

grtjn:
Looks really promising, this would get rid of a lot of unnecessary complexity. Finally ready for running some tests with this. I'll leverage the self-test, and try to run against ML5 up to 8..

grtjn:
Discovered issue #520 when running self-test. Otherwise this runs clean against ML5 up to ML8, and my ML8 cluster!

Merging now, sorry for the delay!

alencruz:
Awesome, thanks!

Issue #511, opened 2015-09-09
xdmp:set-response-code does not seem to work with ML 8.0-3


heelix:
Seems like xdmp:set-response-code does not do anything anymore.  Exceptions (real ones) seem to get forwarded, but if I try to manually set a response code, it still comes back as 200/OK.

declare 
%roxy:params("")
function cs:delete(
    $context as map:map,
    $params  as map:map
) as document-node()?
{
  map:put($context, "output-types", "application/xml"),
 xdmp:set-response-code(501, "Not Implemented"),
  document { "DELETE called" }
};

I see the docs for the ML REST client show something like this now. 

http://docs.marklogic.com/guide/rest-dev/extensions#id_31884

declare function example:put(
    $context as map:map,
    $params  as map:map,
    $input   as document-node()*
) as document-node()?
{
    ...
    map:put($context, "output-status", (201, "Created"));
};

dmcassel:
Are you saying that the output-status part does work but xdmp:set-response-code() does not?

heelix:
Yup.  You see 'DELETE called', with 200/OK rather than the 501/Not Implemented.

dmcassel:
Okay. Looks like our extension template needs an update.

heelix:
That is enough to close the issue with ML8.0-3.  Probably want to have something more with fn:error() later on rather than xdmp:set-response-code... but... xdmp:set-response-code will indeed set the code/message.

PR #512, opened 2015-09-10
#511 corrected extension template

dmcassel:
Response codes should be set in $context, not with xdmp:set-response-code().
http://docs.marklogic.com/guide/rest-dev/extensions#id_31884 #511

### DIFF
--- a/deploy/sample/rest-ext.sample.xqy
+++ b/deploy/sample/rest-ext.sample.xqy
@@ -5,8 +5,8 @@ module namespace yourNSAlias = "http://marklogic.com/rest-api/resource/extension
 declare namespace roxy = "http://marklogic.com/roxy";
 declare namespace rapi = "http://marklogic.com/rest-api";
 
-(: 
- : To add parameters to the functions, specify them in the params annotations. 
+(:
+ : To add parameters to the functions, specify them in the params annotations.
  : Example
  :   declare %roxy:params("uri=xs:string", "priority=xs:int") yourNSAlias:get(...)
  : This means that the get function will take two parameters, a string and an int.
@@ -22,7 +22,7 @@ function yourNSAlias:get(
 ) as document-node()*
 {
   map:put($context, "output-types", "application/xml"),
-  xdmp:set-response-code(200, "OK"),
+  map:put($context, "output-status", (200, "OK")),
   document { "GET called on the ext service extension" }
 };
 
@@ -37,7 +37,7 @@ function yourNSAlias:put(
 ) as document-node()?
 {
   map:put($context, "output-types", "application/xml"),
-  xdmp:set-response-code(200, "OK"),
+  map:put($context, "output-status", (201, "Created")),
   document { "PUT called on the ext service extension" }
 };
 
@@ -53,7 +53,7 @@ function yourNSAlias:post(
 ) as document-node()*
 {
   map:put($context, "output-types", "application/xml"),
-  xdmp:set-response-code(200, "OK"),
+  map:put($context, "output-status", (201, "Created")),
   document { "POST called on the ext service extension" }
 };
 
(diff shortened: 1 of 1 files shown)


grtjn:
Maybe also give an example of how to return an error status? I thought you cannot use output-status for that: http://docs.marklogic.com/guide/rest-dev/extensions#id_33892

dmcassel:
Right, output-status is not for errors. What do you think about a comment up top providing an example and referring to the appropriate documentation?

dmcassel:
Updated as discussed.

deploy/sample/rest-ext.sample.xqy
```diff
@@ -5,11 +5,19 @@ module namespace yourNSAlias = "http://marklogic.com/rest-api/resource/extension
...
-(: 
- : To add parameters to the functions, specify them in the params annotations. 
+(:
+ : To add parameters to the functions, specify them in the params annotations.
  : Example
  :   declare %roxy:params("uri=xs:string", "priority=xs:int") yourNSAlias:get(...)
  : This means that the get function will take two parameters, a string and an int.
+ :
+ : To report errors in your extension, use fn:error(). For details, see
+ : http://docs.marklogic.com/6.0/guide/rest-dev/transforms#id_50497, but here's
```
grtjn: Interesting. You are pointing to 6.0 docs, and to a section related to REST transforms? :)

grtjn: This seems the link to the current REST extension equivalent. Looks slightly different? http://docs.marklogic.com/guide/rest-dev/extensions#id_33892

heelix:
For what it is worth, if you specify a "output-status", and return the fn:error in whatever lib is getting invoked, the error will override whatever was specified in the REST endpoint (/rest-api/ext) code.  If one specified a default status code other than the 200/OK, it would not stick.

(Ended up deleting the rest endpoints we did not want to support, which also kicks a good enough error message to the REST consumer.)

grtjn:
That is what I would have expected indeed. The fn:error is pretty strong, it is like throwing an Exception in Java, and the function will not have the chance to return the result.

Not implementing particular methods returns a 405, method not allowed?

heelix:
400 Bad Request

{
errorResponse: {
statusCode: 400
status: "Bad Request"
messageCode: "XDMP-UNDFUN"
message: "XDMP-UNDFUN: (err:XPST0017) xdmp:function(fn:QName("http://marklogic.com/rest-api/resource/xxx","put"), "/marklogic.rest.resource/deliver/assets/xxx.xqy")($context, $service-params, $input) -- Undefined function put()"
}-
}

grtjn:
Sounds like an RFE against MarkLogic.. :)

deploy/sample/rest-ext.sample.xqy
dmcassel: Doh. I was checking how far back some of the stuff goes, got the wrong link. Will fix.

deploy/sample/rest-ext.sample.xqy
```diff
@@ -5,11 +5,20 @@ module namespace yourNSAlias = "http://marklogic.com/rest-api/resource/extension
...
-(: 
- : To add parameters to the functions, specify them in the params annotations. 
+(:
+ : To add parameters to the functions, specify them in the params annotations.
  : Example
  :   declare %roxy:params("uri=xs:string", "priority=xs:int") yourNSAlias:get(...)
  : This means that the get function will take two parameters, a string and an int.
+ :
+ : To report errors in your extension, use fn:error(). For details, see
+ : http://docs.marklogic.com/guide/rest-dev/transforms#id_50497, but here's
```
grtjn: Still pointing to transforms section? Perhaps use this: http://docs.marklogic.com/guide/rest-dev/extensions#id_33892

dmcassel:
@grtjn got the right link this time

grtjn:
:)

PR #523, opened 2015-09-21
Fixed #508: fixed MLCP options_file support

### LINKED ISSUE #508: var expansion in mlcp options file not working (properly)
Roxy mistakenly looks for `-option_file` instead of `-options_file`. It also looks in `deploy/` for options files which is not intuitive..

grtjn:
#508


dmcassel:
@grtjn Need to update -option_file to -options_file in deploy/Help.rb also.

grtjn:
Like this?

PR #526, opened 2015-09-21
Fixed #503: leveraged json:array to fix helper function

### LINKED ISSUE #503: helper:assert-throws-error, xdmp:apply, varargs
I have a function I want to test.  It takes 6 parameters.  If I call `helper:assert-throws-error`, I get a “Too few args” error message.  The error is actually occurring at line [295](https://github.com/marklogic/roxy/blob/master/src/test/test-helper.xqy#L295).

I think this might be a bug because xdmp:apply seems to require the $params be itemized.  Instead of this

```xquery
	let $params := (‘1’, ‘2’, ‘3’)
	return xdmp:apply( $func, $params )
```
You seem to have to do this:

```xquery
	let $params := (‘1’, ‘2’, ‘3’)
	return xdmp:apply($func, $params[1], $params[2], $params[3])
```

(Side note:  Some googling turned up this [2011 post](https://www.mail-archive.com/general@developer.marklogic.com/msg07547.html) discussing the ‘varargs’ aspect of `xdmp:apply`.)

I’ve actually implemented a fix locally and it seems to work.  Of course, it has all ugliness mentioned in the above post.  Happy to submit a pull request if, in fact, this is an issue.

If it isn’t an issue, please advise how one uses `helper:assert-throws-error` with multiple vars.

grtjn:
#503 

This should do the trick, but want to check backwards compat against ML6 and ML5. I might rewrite to use map:map, which should be much more difficult..

### DIFF
--- a/src/test/test-helper.xqy
+++ b/src/test/test-helper.xqy
@@ -283,16 +283,60 @@ declare function helper:assert-throws-error($function as xdmp:function)
 
 declare function helper:assert-throws-error($function as xdmp:function, $error-code as xs:string?)
 {
-  helper:assert-throws-error($function, "no-params-please", $error-code)
+  helper:assert-throws-error($function, (), $error-code)
 };
 
-declare function helper:assert-throws-error($function as xdmp:function, $params as item()*, $error-code as xs:string?)
+declare function helper:assert-throws-error($function as xdmp:function, $param1 as item()*, $error-code as xs:string?)
 {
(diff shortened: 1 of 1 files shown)


grtjn:
According to DMC json:to-array exists in ML6, but not in ML5. Good enough?

dmcassel:
ML5 is already past its end-of-life date, so yeah, we don't have to support it anymore.

Issue #500, opened 2015-07-19
Don't add schedules to all hosts by default


grtjn:
People might not be aware that schedules could be running on multiple hosts in a group/cluster, add bootstrap host as default in sample config. Removing it would give old behavior..

grtjn:
Ok, I did an attempt by adding `<task-host name="@ml.server"/>` to `ml-config.sample.xml`, but @dmcassel pointed correctly out to me that what you type in as server in your env properties, doesn't have to match with the real host name inside MarkLogic. We need something that refers to hosts by their ML host name.

- `@ml.server-hostname`
- `@ml.host-names[1]`
- `{xdmp:host-name()}` (evaluated in setup.xqy, not very safe?)
- `{host-name}` (interpreted/replaced in setup.xqy)

 Anyone better ideas?

grtjn:
Needs more discussion, pushing away..

grtjn:
I vote for `${server-name}`, or `@ml.server-name` in old syntax. The value will need to be derived dynamically by executing `xdmp:host-name(xdmp:host())`..

grtjn:
Fixed in DEV

PR #527, opened 2015-09-22
Fixed #497/#500: fixed reading task-host, changed its default

### LINKED ISSUE #497: adding scheduled task to ml-config causes lexval error on bootstrap
I have a configuration with multiple hosts, divided over two groups.
I added a scheduled task to the server section in a specific group in ml-config and added a host-name where the task should run. When bootstrapping I get the following error:
ml local bootstrap
Bootstrapping your project into MarkLogic 8 on ln-vm4-ml2...
ERROR: XDMP-LEXVAL: gr:host-id("") -- Invalid lexical value ""
See MarkLogic Server error log for more details.
ERROR: ... Bootstrap FAILED

If I change the way the host-name is handled in the function setup:create-scheduled-task
from this:
        $task/gr:task-host/xdmp:host(.),
to this:
        xdmp:host($task/gr:task-host/@name),
and change the way it is configured in ml-config from element content to attribute name, bootstrap will work...

grtjn:
Self-test against ML8 runs clean.


dmcassel:
"@ml.server" may be "localhost", which is good enough to direct traffic to MarkLogic, but xdmp:host("localhost") will fail. 

> XDMP-NOSUCHHOST: xdmp:host("localhost") -- No such host localhost

grtjn:
Hmm, good one. We need something smarter, just not sure what. I'd rather not default to bootstrap-host or first host in group, to keep the option to have a schedule on all hosts..

dmcassel:
Localhost is a special case. I think if we'd be safe guessing that "localhost" => xdmp:host().

grtjn:
Yeah, but the assumption is actually all wrong. `@ml.server` doesn't contain a host-name at all, but something that happens to resolve to the MarkLogic host ip against the DNS or the local hosts file..

dmcassel:
Agreed. Sometimes those will match, sometimes not.

peetkes:
Works fine by me, except for the placeholder in the sample config. You should be able to pick any of the hosts in the group in a convenient way `@ml.group-hosts[n]`

grtjn:
I think I'll split this PR. Peet's suggestion could be useful, but needs more thought..

dmcassel:
@grtjn should we close this PR for now?

grtjn:
Not needed. I'll just take out the configs, and continue discussion about it in #500..

dmcassel:
@grtjn what's the state of this PR? Still valid?

grtjn:
Yes, should be ready to merge. I took out the tricky part with host names (#500)..

PR #529, opened 2015-09-22
Fixed #487: added MLCP_HOME/bin to PATH

### LINKED ISSUE #487: Check running mlcp 1.3-3 on windows..
Something with HADOOP_HOME (or better -Dhadoop.home.dir=..) and a not running MLCP without that..

grtjn:
#487 

Anyone around that can run a quick test on Windows? Runs fine on Mac..

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1208,8 +1208,13 @@ In order to proceed please type: #{expected_response}
     logger.debug runme
     logger.info ""
 
-    system runme
-
+    # PATH change only important for Windows, so always using ; and \
+    env_variables = {
+      "PATH" => "#{ENV['PATH']};#{mlcp_home}\\bin", 
+      "HADOOP_HOME" => mlcp_home
+    }
+    system(env_variables, runme)
+    
     logger.info ""
 
     ARGV.clear


patrickmcelwee:
Geert,

I suggest setting HADOOP_HOME too. That was an issue I ran into while working through #542 . I was not able to run a completely successful test on Windows, since I ran into another problem with mlcp 1.3-3 on Windows 10 (again, see #542 ) - but I think that problem was unrelated to this.

Here is how my code looked:

```ruby
    env_variables = {"PATH" => "#{ENV['PATH']};#{mlcp_home}\bin", 
                                "HADOOP_HOME" => mlcp_home}
    system(env_variables, runme)
```

grtjn:
I've added HADOOP_HOME too. Could you run a quick test with this to see if this works on Windows? (i.e. doesn't miserably fail when executing `ml.bat local mlcp`)

patrickmcelwee:
Yep, an import of an XML doc worked for me. Though, it also worked against the current Roxy master. Not sure what that means. Possibly it is working just because I am using the most recent version of mlcp (8.0-4)

grtjn:
If the change didn't help, then it didn't harm either. I think we can safely merge this..

patrickmcelwee:
agreed

PR #530, opened 2015-09-22
Fixed #457: assign forests to group host 1 instead of default-host

### LINKED ISSUE #457: Create database in non-bootstrap hosts can fail
A database that is not using `<forests-per-host>`, with a forest without explicit `<host-name>`, assigned to a group different than the one from the bootstrap host, will fail to create with a vague cast error telling:

```
XDMP-AS: (err:XPTY0004) $hostnr as xs:integer -- Invalid coercion: () as xs:integer
```

The code tries to determine the host for the single forest it is trying to create, but there is not explicit host-name. It then falls back to default-host, which is the bootstrap host. It will then try to determine hostnr against the hosts within the group of the database. But if bootstrap host is in a different group, then this fails.

I think it is better to fall back to the first of the hosts within the group, instead of default-host. That would mean hostnr would be always 1, and these kind of databases (like appname-modules and such) would all get attached to the same host, but that is already the case. You can always use `<host-name>` on the forest assignments to assign a particular forest to a particular host.

grtjn:
#457


peetkes:
Works fine by me, so please accept

PR #532, opened 2015-09-22
Fixed #515: added reading {env}.json

### LINKED ISSUE #515: Read json env properties
Latest changes to slush-marklogic-node have introduced a generic {env}.json properties file to be used by various processes. Would be very convenient if Roxy would use those properties if present for overriding properties. Should be stronger than local.properties, though a warning might be nice if both local.json and deploy/local.properties exist.

Note: the json properties will be in the project root, not inside deploy/..

grtjn:
#515 

This is how reading local.json would look like. Open for debate..

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -23,6 +23,7 @@ require 'xcc'
 require 'MLClient'
 require 'date'
 require 'ml_rest'
+require 'yaml'
 
 class ExitException < Exception; end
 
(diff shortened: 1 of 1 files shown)


grtjn:
Note: I checked, YAML should be a library that is installed by default. At least it exists on demo servers without installing new gems..

dmcassel:
The issue this PR refers to was closed as a do-not-fix, so I'm going to close the PR, too.

Issue #543, opened 2015-10-29
Unable to redirect output from the "ml" script using stanard methods


danielritchie:
## Overview of the Issue
* Executing the script as a user at the terminal works as expected.  However, exception output is not captured when executed programmatically. 
* We discovered this issue when using Jenkins CI, as the same text which is not being redirected is also not showing up in Jenkin's console (unsure if for the same reason, but seems likely)
* We could figure out a way to redirect the output, but making the proposed changes seems a more complete solution

###### Example:
```
# ./ml bootstrap
ERROR: Missing environment for bootstrap
Usage: ml {env} bootstrap [options]

General options:
 -v, [--verbose]  # Verbose output

Bootstraps your application to the MarkLogic server in the given
environment.
```
###### --- VS: ---
```
# ./ml bootstrap > output.log 
# ls -s output.log
0 output.log
```
###### NOTES: 
* No luck with different redirects/tee/etc.
* Setting verbose flag had no effect
* Interestingly, the -h flag does return the output (discussed in more detail under "Suggest a Fix") - UPDATE - Help works because it returns using "exit" instead of "exit!"
```
# ./ml bootstrap -h > output.log
# cat output.log
Usage: ml {env} bootstrap [options]

General options:
  -v, [--verbose]  # Verbose output

Bootstraps your application to the MarkLogic server in the given
environment.
```
## Motivation for or Use Case
* When "ml" is executed programmatically output cannot be captured.  This prevents things like automating deployments and implementing CI.  It also means troubleshooting roxy has become my least favorite thing to do
* Full disclosure: I'm not familiar enough with roxy/ruby to be certain that I'm not overlooking a known setting that would enable the output to be captured...but haven't had any luck finding anything

##  Roxy Version
* 1.7.3-dev
* Also confirmed on 1.7.2

##  Ruby Version
* 1.9.3
* Also confirmed on 1.8.7

## Operating System
* Centos 6.5

###### UPDATE 1:
* **The issue appears to be with the "exit!" statement at line 187 in deploy/lib/ml.rb**
* While this does result in the expected return code of 1 (as I'm sure is intended), it also seems to have an unintended side effect of wiping out the messaging.  
* Commenting out the "exit!" @ line 187 allows the redirect to be captured (I haven't tested with Jenkins yet), but introduces a new problem of returning 0.  
[...]

PR #544, opened 2015-10-30
Fix for issue #543: updated all exit calls with a unique return code

danielritchie:
***deploy/lib/ml.rb*** presently exits with errors using "exit!" This forces an immediate exit that bypasses handling and returns "1".  Using this method also causes something strange to happen with the output of the script, preventing redirection and other concerns [See: Issue #543](https://github.com/marklogic/roxy/issues/543)

**All "exit!" calls have been replaced with "exit #"** as I couldn't see any reason why the use of return codes would be problematic.  I believe this maintains all existing functionality, and is a desirable enhancement.

**There was a single "exit" call which was also replaced with a unique return code, "exit 10"** - Note that this will change exiting behavior ("exit" alone will return 0).  Although I expect this will be a desired addition, I wanted to call it out since it does change the return code.

More detail from Ruby docs:
[exit](http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-exit) vs. [exit!](http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-exit-21)

### DIFF
--- a/deploy/lib/ml.rb
+++ b/deploy/lib/ml.rb
@@ -39,7 +39,7 @@ if @profile then
     RubyProf.start
   rescue LoadError
     print("Error: Please install the ruby-prof gem to enable profiling\n> gem install ruby-prof\n")
-    exit
+    exit 10
   end
 end
 
@@ -125,7 +125,7 @@ begin
         ServerConfig.no_prompt = @no_prompt
         result = ServerConfig.send command
         if !result
-          exit!
+          exit 1
         end
       end
       break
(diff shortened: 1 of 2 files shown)


danielritchie:
With unique return codes being available, it seemed worthwhile to pass them back to the parent script. 

###### Previously:
```
# ./ml bootstrap
ERROR MESSAGE
# echo $?
1
```

###### With this change:
```
# ./ml bootstrap
ERROR MESSAGE
# echo $?
7
```

danielritchie:
Just realized I need to submit an updated pull request to the dev branch...will do so in the AM

danielritchie:
Updated pull request: #546

PR #546, opened 2015-10-31
Capture/redirect on exit and provide unique return codes, fix for issue #543

danielritchie:
### deploy/lib/ml.rb
***deploy/lib/ml.rb*** presently exits with errors using "exit!" This forces an immediate exit that bypasses [SystemExit](http://ruby-doc.org/core-1.9.3/SystemExit.html) handling and returns "1".  Using this method also causes something strange to happen with the output of the script, preventing redirection and other concerns [See: Issue #543](https://github.com/marklogic/roxy/issues/543)

**All "exit!" calls have been replaced with "exit #"** as I couldn't see any reason why the use of return codes would be problematic.  I believe this maintains all existing functionality, and is a desirable enhancement.

**Two "exit" calls were also replaced with a unique return code** THIS WILL CHANGE EXIT BEHAVIOR ("exit" alone will return 0).  Although I don't see any reason this addition would be undesired, I wanted to call it out since it does modify the return code.

More detail from Ruby docs:
[exit](http://ruby-doc.org/core-1.9.3/Kernel.html#method-i-exit) vs. [exit!](http://ruby-doc.org/core-1.9.3/Kernel.html#method-i-exit-21)
[SystemExit](http://ruby-doc.org/core-1.9.3/SystemExit.html)

###  ml
**ml** has been modified to exit with actual return code from deploy/lib/ml.rb instead of defaulting to 1

### DIFF
--- a/deploy/lib/ml.rb
+++ b/deploy/lib/ml.rb
@@ -39,7 +39,7 @@ if @profile then
     RubyProf.start
   rescue LoadError
     print("Error: Please install the ruby-prof gem to enable profiling\n> gem install ruby-prof\n")
-    exit
+    exit 1
   end
 end
 
@@ -54,7 +54,7 @@ end
 
 if ARGV.length == 1 && need_help?
   Help.doHelp(@logger, :usage)
-  exit
+  exit 2
 end
 
 if RUBY_VERSION < "1.8.7"
@@ -125,7 +125,7 @@ begin
         ServerConfig.no_prompt = @no_prompt
         result = ServerConfig.send command
         if !result
-          exit!
+          exit 3
         end
       end
       break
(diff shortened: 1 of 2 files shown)


danielritchie:
Have confirmed that when using this fix the output is returned as expected when executing from Jenkins

dmcassel:
Regarding the two exit calls replaced with exit and a return code -- that looks like a good thing to me, as exit code 0 is typically success.

dmcassel:
I ran some tests and this looks good. One thing that might be useful as a future change is replace the literal numbers with constant values, to make it easier to look up the error codes. 

@danielritchie thanks for submitting this!

danielritchie:
@dmcassel your wish has been at least partially graned with #547 :)

PR #547, opened 2015-11-03
PREP to use constanst for exit codes

danielritchie:
Follow up on comment by @dmcassel when working through https://github.com/marklogic/roxy/pull/546

It was easy for me with everything open to perform a few quick substitutions.  Past that, I attempted to confirm Ruby syntax but wasn't 100% sure I got it right.  I also don't have a good environment setup for testing this.  Having said that, if it doesn't work as-is, the placeholders are staged so it should be ready to go for any changes you would like to make.

  - replaced all hard-coded exit codes with constants
  - set usage to return 0 if help requested
  - shifted codes to avoided using reserved bash return codes

### DIFF
--- a/deploy/lib/ml.rb
+++ b/deploy/lib/ml.rb
@@ -20,6 +20,19 @@ require 'util'
 require 'upgrader'
 require 'scaffold'
 
+#Return Codes
+EXIT_LOAD_PROF_GEM = 3
+EXIT_SHOW_HELP = 0
+EXIT_SERVER_CONFIG_SEND = 4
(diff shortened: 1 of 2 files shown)


paxtonhare:
@danielritchie can you take off that out.test file?

danielritchie:
@paxtonhare done

dmcassel:
Minor typo: could you replace EXIT_RESCUE_EXCPTION with EXIT_RESCUE_EXCEPTION?

danielritchie:
@dmcassel var name replaced as requested

PR #552, opened 2015-11-06
Fix #551: Only load content that is newer locally than on the server

### LINKED ISSUE #551: Deploy only newer content
We have a change locally that allows the passing of a command line flag to "ml deploy content" so that it will only deploy files which are new or have a more recent modification date than the equivalent on the server (using document timestamp). We have found this useful when the deployed content is a set of "baseline" documents that can be modified in a running system.

Is this change something that is likely to be merged if turned into a Pull Request?

Thanks

Chris

nespera:
This change adds support for a command line flag that causes "deploy content" to compare modification times on local files and timestamps on deployed documents. It then only deploys files which are new or have newer modification dates locally.

The use case is the situation where the deployed files are a "baseline" set that could potentially change in a running system.

~~It's not optimized for speed and will be slow if there's a large number of small files, as an individual timestamp query is done for each.~~

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1387,12 +1387,13 @@ private
     end
 
     uris = files.map { |f| xcc.build_target_uri(f, options) }
+    uris_as_string = uris.map{|i| "\"#{i}\""}.join(",")
+    q = %Q{for $u in (#{uris_as_string}) return "" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp($u))}
 
-    stamps_db = uris.map do |target_uri|
-      q = %Q{"" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp("#{target_uri}"))}
-      result = execute_query q, :db_name => @properties["ml.content-db"]
-      parse_json(result.body)
-    end
+    result = execute_query q, :db_name => @properties["ml.content-db"]
+    stamps_db = parse_json(result.body).split("\n")
+
+    print stamps_db.class
 
     stamps_local = files.map { |file_uri| File.mtime(file_uri).getgm.iso8601(5).tr('Z','') }
 
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1387,14 +1387,7 @@ private
     end
 
     uris = files.map { |f| xcc.build_target_uri(f, options) }
-    uris_as_string = uris.map{|i| "\"#{i}\""}.join(",")
-    q = %Q{for $u in (#{uris_as_string}) return "" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp($u))}
-
-    result = execute_query q, :db_name => @properties["ml.content-db"]
-    stamps_db = parse_json(result.body).split("\n")
-
-    print stamps_db.class
-
+    stamps_db = get_db_timestamps(uris)
     stamps_local = files.map { |file_uri| File.mtime(file_uri).getgm.iso8601(5).tr('Z','') }
 
     files_with_stamps = files.zip(stamps_local, stamps_db)
@@ -1411,6 +1404,14 @@ private
     end
 
     filtered.map { |f, stamp1, stamp2| f}
+  end
+
+  def get_db_timestamps(uris)
+    uris_as_string = uris.map{|i| "\"#{i}\""}.join(",")
+    q = %Q{for $u in (#{uris_as_string}) return "" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp($u))}
+
+    result = execute_query q, :db_name => @properties["ml.content-db"]
+    parse_json(result.body).split("\n")
   end
 
   def save_files_to_fs(target_db, target_dir)
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1386,14 +1386,19 @@ private
       raise ExitException.new("Can only filter files on MarkLogic 6 and later")
     end
 
-    files.select { |file_uri|
-      target_uri = xcc.build_target_uri(file_uri, options)
+    uris = files.map { |f| xcc.build_target_uri(f, options) }
 
+    stamps_db = uris.map do |target_uri|
       q = %Q{"" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp("#{target_uri}"))}
       result = execute_query q, :db_name => @properties["ml.content-db"]
-      stamp_in_db = parse_json(result.body)
+      parse_json(result.body)
+    end
 
-      stamp_locally = File.mtime(file_uri).getgm.iso8601(5).tr('Z','')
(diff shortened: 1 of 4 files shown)


deploy/lib/server_config.rb
```diff
@@ -1357,6 +1374,28 @@ def settings
 
 private
 
+  def filter_to_newer_files(files, options)
+    logger.info "Filtering to files which are newer locally than on the server"
+
+    files.select { |file_uri|
+      target_uri = xcc.build_target_uri(file_uri, options)
+
+      q = %Q{"" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp("#{target_uri}"))}
```
grtjn: Interesting, leverages xdmp:document-timestamp! Won't work on ML5, but does on 6+.

deploy/lib/server_config.rb
```diff
@@ -1357,6 +1374,28 @@ def settings
...
+      result = execute_query q, :db_name => @properties["ml.content-db"]
```
grtjn: Wondering if it would be more efficient to do a single query for all uris (returned as some hash?), or perhaps calls in batches, instead of calling ML for each uri separately. Ideally it would be that fast that it becomes worth keeping it enabled permanently..

deploy/lib/util.rb
```diff
@@ -102,6 +102,36 @@ def file_exists?(filename)
...
+        get_files(full_path, options, data)
+      else
+        data << full_path.encode("UTF-8")
+      end
+    end
+  else
+    data = [path.encode("UTF-8")]
+  end
+  data
+end
```
grtjn: You moved this file to util, but isn't it possible to just push the timestamp filter code inside the xcc library? Or do you see potential reuse for other commands, not using xcc?

nespera: I could potentially pass the list of files and timestamps to xcc and do the filtering there. I moved this method as it seemed strange to call into xcc to do something only concerned with local files.

deploy/lib/server_config.rb
nespera: Yes, I suspect it would be faster that way.

deploy/lib/server_config.rb
nespera: Maybe I should add some error handling so it will not break on ML5?

deploy/lib/util.rb
nespera: Actually, I'm using execute_query inside server_config to get the timestamps, so that code can't easily move inside xcc. In order to get the timestamps I need the list of files, so that also needs to be available to server_config. As I said, I could leave that in xcc, but it seemed more natural to put that in util with other filesystem utility methods.

nespera:
I have now changed it so that it will:
- give an error if you try to filter before ML6
- hit ML once to determine all the timestamps

I thought it could be an issue if you have an *enormous* list of files, but I simulated the XQuery call in qconsole and it was happy checking 10000 document timestamps at once.

deploy/lib/server_config.rb
```diff
@@ -1362,6 +1379,41 @@ def settings
...
+    end
+
+    filtered.map { |f, stamp1, stamp2| f}
+  end
+
+  def get_db_timestamps(uris)
+    uris_as_string = uris.map{|i| "\"#{i}\""}.join(",")
+    q = %Q{for $u in (#{uris_as_string}) return "" || xdmp:timestamp-to-wallclock(xdmp:document-timestamp($u))}
+
+    result = execute_query q, :db_name => @properties["ml.content-db"]
```
dmcassel: Your query is running against the content database. I didn't get any timestamps back until I changed this to ["ml.modules-db"].

deploy/lib/server_config.rb
```diff
@@ -1362,6 +1379,41 @@ def settings
...
+  def filter_to_newer_files(files, options)
+    logger.info "Filtering to files which are newer locally than on the server"
+
+    if @server_version < 6
+      raise ExitException.new("Can only filter files on MarkLogic 6 and later")
+    end
+
+    uris = files.map { |f| xcc.build_target_uri(f, options) }
+    stamps_db = get_db_timestamps(uris)
+    stamps_local = files.map { |file_uri| File.mtime(file_uri).getgm.iso8601(5).tr('Z','') }
```
dmcassel: The local timestamps are accounting for timezone, but get_db_timestamps() doesn't. In my case, that meant my local timestamps were 5 hours ahead, so all files needed to be loaded. After removing the `.getgm` and the `.tr(...)` from the line above, Ruby compared timestamps that looked like:

- local: 2015-12-17T13:13:30.00000-05:00
- in db: 2015-12-17T13:10:00.229146

While they look slightly different, they are in the same timezone.

To make this work for remote servers, I think you'll need to call [fn:implicit_timezone()](http://docs.marklogic.com/fn:implicit-timezone) on the MarkLogic server, then use that value with [Time.getlocal(utc_offset)](http://ruby-doc.org/core-2.2.0/Time.html#method-i-getlocal) for the local times.

deploy/lib/server_config.rb
nespera: This function is looking at timestamps on documents in the content database, so isn't that the correct database to be running against?

deploy/lib/server_config.rb
nespera: I've now pushed a change to ensure that both the server timestamps and the local ones are in UTC. One of the joys of being UK-based is that it's easy to make timezone screw ups without noticing since half the year we are on UTC...

deploy/lib/server_config.rb
dmcassel: Is this being used to deploy content (data) or modules (code)? If you're deploying modules, they are sent to the modules database, so you'll need to check the timestamps there.

nespera: It is only used for "deploy content"

dmcassel: Ah, my mistake.

grtjn: I think this feature would be very much appreciated for deploy modules as well, at the least. For that it would be useful if it would be more dynamic..

nespera: I could potentially pass in the database name to get_db_timestamps, but it seems like unneeded flexibility at this point. Having incremental deployment of modules doesn't seem useful to me so I will leave that (and making get_db_timestamps more flexible) to someone else.

deploy/lib/util.rb
dmcassel: @grtjn are you satisfied with this response?

grtjn: Yeah, I guess so. I was only talking about get_files itself that stands on its own, and seems to have moved unchanged from xcc to util (except for IGNORE_EXTENSIONS). I'm fine with it provided we test this properly, and it doesn't break anything. :)

dmcassel:
Looks like what was identified previously has been fixed.

PR #577, opened 2016-03-04
Issue 109

dmcassel:
Added commands:

- ml {env} deploy triggers
- ml {env} clean triggers

The code tests to make sure you have a Triggers database configured. The clean wipes out all triggers in the target database, so users are advised to have an application-specific triggers database.

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1878,16 +1878,19 @@ private
       raise ExitException.new("Deploy triggers requires a triggers database")
     end
 
-    if !File.exist?(ServerConfig.expand_path("#{@@path}/triggers-config.xml"))
+    target_config = ServerConfig.expand_path(ServerConfig.properties["ml.triggers.file"])
+
+    if !File.exist?(target_config)
       logger.error <<-ERR.strip_heredoc
         Before you can deploy triggers, you must define a configuration. Steps:
-        1. Copy deploy/sample/triggers-config.sample.xml to deploy/triggers-config.xml
-        2. Edit deploy/triggers-config.xml to specify your trigger(s)
+        1. Copy deploy/sample/triggers-config.sample.xml to #{target_config}
+          The location of this file is controlled by the triggers.file property.
+        2. Edit #{target_config} to specify your trigger(s)
         3. Run 'ml <env> deploy triggers')
       ERR
     else
-      triggers_config = File.read ServerConfig.expand_path("#{@@path}/triggers-config.xml")
-      replace_properties(triggers_config, "triggers-config.xml")
+      triggers_config = File.read target_config
+      replace_properties(triggers_config, target_config)
       triggers_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
       query = %Q{#{triggers_code} triggers:load-from-config(#{triggers_config})}
       logger.debug(query)
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1008,6 +1008,8 @@ In order to proceed please type: #{expected_response}
         clean_schemas
       when 'cpf'
         clean_cpf
+      when 'triggers'
+        clean_triggers
       else
         raise HelpException.new("clean", "Invalid WHAT")
     end
@@ -1889,8 +1891,13 @@ private
       triggers_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
       query = %Q{#{triggers_code} triggers:load-from-config(#{triggers_config})}
       logger.debug(query)
-      r = execute_query(query, :db_name => @properties["ml.content-db"])
+      r = execute_query(query, :db_name => @properties["ml.triggers-db"])
     end
+  end
+
+  def clean_triggers
+    cpf_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
+    r = execute_query %Q{#{cpf_code} cpf:clean-triggers()}, :db_name => @properties["ml.triggers-db"]
   end
 
   def xcc
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -937,6 +937,8 @@ In order to proceed please type: #{expected_response}
         deploy_schemas
       when 'cpf'
         deploy_cpf
+      when 'triggers'
+        deploy_triggers
       else
         raise HelpException.new("deploy", "Invalid WHAT")
     end
@@ -1866,6 +1868,29 @@ private
   def clean_cpf
     cpf_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/cpf.xqy")
     r = execute_query %Q{#{cpf_code} cpf:clean-cpf()}, :db_name => @properties["ml.content-db"]
+  end
+
+  def deploy_triggers
+    logger.info "Deploying Triggers"
+    if !@properties["ml.triggers-db"]
+      raise ExitException.new("Deploy triggers requires a triggers database")
+    end
+
+    if !File.exist?(ServerConfig.expand_path("#{@@path}/triggers-config.xml"))
+      logger.error <<-ERR.strip_heredoc
+        Before you can deploy triggers, you must define a configuration. Steps:
+        1. Copy deploy/sample/triggers-config.sample.xml to deploy/triggers-config.xml
+        2. Edit deploy/triggers-config.xml to specify your trigger(s)
+        3. Run 'ml <env> deploy triggers')
+      ERR
+    else
+      triggers_config = File.read ServerConfig.expand_path("#{@@path}/triggers-config.xml")
+      replace_properties(triggers_config, "triggers-config.xml")
+      triggers_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
+      query = %Q{#{triggers_code} triggers:load-from-config(#{triggers_config})}
+      logger.debug(query)
+      r = execute_query(query, :db_name => @properties["ml.content-db"])
+    end
   end
 
   def xcc
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1407,6 +1407,7 @@ Provides listings of various kinds of settings supported within ml-config.xml.
     target_config = ServerConfig.expand_path(ServerConfig.properties["ml.triggers.file"])
 
     if !File.exist?(target_config)
+      logger.error "ml.triggers.file=#{ServerConfig.properties['ml.triggers.file']}"
       logger.error <<-ERR.strip_heredoc
         Before you can deploy triggers, you must define a configuration. Steps:
         1. Copy deploy/sample/triggers-config.sample.xml to #{target_config}
@@ -1422,6 +1423,7 @@ Provides listings of various kinds of settings supported within ml-config.xml.
       query = %Q{#{triggers_code} triggers:load-from-config(#{triggers_config})}
       logger.debug(query)
       r = execute_query(query, :db_name => @properties["ml.triggers-db"])
+      logger.info "... triggers deployed"
       return true
     end
   end
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1896,8 +1896,8 @@ private
   end
 
   def clean_triggers
-    cpf_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
-    r = execute_query %Q{#{cpf_code} cpf:clean-triggers()}, :db_name => @properties["ml.triggers-db"]
+    triggers_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
+    r = execute_query %Q{#{triggers_code} triggers:clean-triggers()}, :db_name => @properties["ml.triggers-db"]
   end
 
   def xcc
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1002,8 +1002,6 @@ In order to proceed please type: #{expected_response}
         clean_content
       when 'modules'
         clean_modules
-      when 'triggers'
-        clean_triggers
       when 'schemas'
         clean_schemas
       when 'cpf'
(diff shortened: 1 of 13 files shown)


deploy/lib/server_config.rb
```diff
@@ -1868,6 +1872,34 @@ def clean_cpf
...
+      raise ExitException.new("Deploy triggers requires a triggers database")
+    end
+
+    if !File.exist?(ServerConfig.expand_path("#{@@path}/triggers-config.xml"))
+      logger.error <<-ERR.strip_heredoc
+        Before you can deploy triggers, you must define a configuration. Steps:
+        1. Copy deploy/sample/triggers-config.sample.xml to deploy/triggers-config.xml
+        2. Edit deploy/triggers-config.xml to specify your trigger(s)
+        3. Run 'ml <env> deploy triggers')
+      ERR
```
grtjn: Nice error handling..

deploy/lib/server_config.rb
```diff
@@ -1868,6 +1872,34 @@ def clean_cpf
     r = execute_query %Q{#{cpf_code} cpf:clean-cpf()}, :db_name => @properties["ml.content-db"]
   end
 
+  def deploy_triggers
+    logger.info "Deploying Triggers"
+    if !@properties["ml.triggers-db"]
+      raise ExitException.new("Deploy triggers requires a triggers database")
+    end
+
+    if !File.exist?(ServerConfig.expand_path("#{@@path}/triggers-config.xml"))
```
grtjn: How about making `triggers-config.xml` configurable via properties, like `ml-config.xml`?

deploy/lib/server_config.rb
```diff
@@ -1868,6 +1872,34 @@ def clean_cpf
...
+        2. Edit deploy/triggers-config.xml to specify your trigger(s)
+        3. Run 'ml <env> deploy triggers')
+      ERR
+    else
+      triggers_config = File.read ServerConfig.expand_path("#{@@path}/triggers-config.xml")
+      replace_properties(triggers_config, "triggers-config.xml")
+      triggers_code = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/triggers.xqy")
+      query = %Q{#{triggers_code} triggers:load-from-config(#{triggers_config})}
+      logger.debug(query)
+      r = execute_query(query, :db_name => @properties["ml.triggers-db"])
```
grtjn: I like the fact that you pushed the triggers code into a separate .xqy..

deploy/lib/xquery/triggers.xqy
```diff
@@ -0,0 +1,97 @@
...
+ :@param $config - the configuration xml
+ :
+:)
+declare function triggers:load-from-config($config as element(trgr:triggers))
+{
+  let $triggers-db := xdmp:database()
+  for $trgr in $config/trgr:trigger
+  let $name := $trgr/trgr:name
+  let $desc as xs:string := $trgr/trgr:description
+  let $data-event := $trgr/trgr:data-event
```
grtjn: Would it be difficult to add support for database-online-event?

deploy/lib/xquery/triggers.xqy
```diff
@@ -0,0 +1,97 @@
...
+      }
+    </trgr:module>
+  let $enabled as xs:boolean := $trgr/trgr:enabled
+  let $permissions as element()* :=
+    (
+      triggers:resolve-permissions($trgr/trgr:permissions/sec:permission),
+      xdmp:default-permissions()
+    )[1]
+  let $recursive as xs:boolean := ($trgr/trgr:recursive, fn:true())[1]
+  let $priority as xs:string? := $trgr/trgr:task-priority
```
grtjn: I think it is safer to do the `(.., "normal")[1]` here..

deploy/lib/xquery/triggers.xqy
```diff
@@ -0,0 +1,97 @@
...
+ :
+ :@param $config - the configuration xml
+ :
+:)
+declare function triggers:load-from-config($config as element(trgr:triggers))
+{
+  let $triggers-db := xdmp:database()
+  for $trgr in $config/trgr:trigger
+  let $name := $trgr/trgr:name
+  let $desc as xs:string := $trgr/trgr:description
```
grtjn: desc is optional..

deploy/lib/xquery/triggers.xqy
```diff
@@ -0,0 +1,97 @@
...
+  let $data-event := $trgr/trgr:data-event
+  let $module :=
+    <trgr:module>
+      <trgr:database>{xdmp:database($trgr/trgr:module/trgr:database/fn:string())}</trgr:database>
+      {
+        $trgr/trgr:module/* except $trgr/trgr:module/trgr:database
+      }
+    </trgr:module>
```
grtjn: How about using http://docs.marklogic.com/trgr:trigger-module here?

deploy/lib/xquery/triggers.xqy
```diff
@@ -0,0 +1,97 @@
...
+};
+
+declare function triggers:resolve-permissions($perms as element(sec:permission)*)
+{
+  for $perm in $perms
+  return
+    <sec:permission>
+      {$perm/sec:capability}
+      <sec:role-id>{xdmp:role($perm/sec:role-name)}</sec:role-id>
+    </sec:permission>
```
grtjn: How about using http://docs.marklogic.com/xdmp:permission here?

deploy/lib/xquery/triggers.xqy
```diff
@@ -0,0 +1,97 @@
...
+};
+
+declare function triggers:clean-triggers()
+{
+  for $trigger in /trgr:trigger
+  return
+    xdmp:document-delete(fn:base-uri($trigger))
```
grtjn: Was expecting a clear-forest here, but I actually like this better. Handy in case you mingle triggers with something else, like schemas or modules.. :)

deploy/sample/triggers-config.sample.xml
```diff
@@ -0,0 +1,34 @@
...
+      </sec:permission>,
+      <sec:permission>
+        <sec:capability>update</sec:capability>
+        <sec:role-name>${app-role}</sec:role-name>
+      </sec:permission>
+    </trgr:permissions>
+    <trgr:recursive>false</trgr:recursive>
+    <trgr:task-priority>normal</trgr:task-priority>
+  </trgr:trigger>
+</trgr:triggers>
```
grtjn: Can we integrate this into self-test?

grtjn:
Tidy work! Above a few inline comments..

deploy/lib/xquery/triggers.xqy
dmcassel: I missed that, good catch

deploy/lib/server_config.rb
dmcassel: Good idea

deploy/lib/xquery/triggers.xqy
dmcassel: The config produces the same XML as trgr:trigger-module(), except that the config takes a database name and the function wants a database id. Compare with data-event, where I just take exactly what's in the config and pass that to create-trigger(). Seems redundant load up the expected XML, then parse out pieces and pass them to a function to get the same thing back. Let me know if you feel strongly about this and I'll change it.

deploy/lib/xquery/triggers.xqy
dmcassel: As with above.

deploy/lib/xquery/triggers.xqy
dmcassel: In a perfect world, they shouldn't be mixed, but I figured this was a little safer.

deploy/sample/triggers-config.sample.xml
dmcassel: I could use some help with that.

deploy/lib/xquery/triggers.xqy
grtjn: I mention recrearing the xml with built-in functions because humans can put in things that cause trouble down the line (we have had that with indexes), and the built-in xml format could change (like addition of new elements). Not highly likely does make it more robust i think..

deploy/sample/triggers-config.sample.xml
grtjn: Let me give self-testing a bit of thought. Could perhaps be added in between bootstrap/validate of ml7-config and the changed one..

deploy/lib/xquery/triggers.xqy
dmcassel: Upon reflection, I agree. I just pushed an update for this.

deploy/lib/xquery/triggers.xqy
dmcassel: Updated this too.

grtjn:
Understood. Integrating this into self-test now makes it easier for all to test this, though. I only looked at the code so far, haven't ran it yet.

I had a quick thought on this, and was thinking you might get away with it by adding to extra test steps to the bootstrap test case. Somewhere around here:

https://github.com/marklogic/roxy/blob/master/deploy/test/test_server_config.rb#L101

Copy the trigger config into test/data/, point to it in the step, and run deploy_triggers. A quick count on trigger docs in the triggers db could verify success/fail. the second step would run clean_triggers, and the count should return zero. Nothing too fancy, just make sure the code runs without errors, and something gets created/cleaned..

Doable?

grtjn:
Self-test doesn't run out of the box. I think some lookup fails if you don't run against a clustered environment. It is relatively easy to make it work though, just edit this line:

https://github.com/marklogic/roxy/blob/dev/deploy/lib/xquery/setup.xqy#L1108

Replace `$hosts[1]` with `($hosts, $default-host)[1]`.

You may also want to comment this line:

https://github.com/marklogic/roxy/blob/dev/deploy/lib/xquery/setup.xqy#L5589

It produces a lot of unnecessary comments in the ErrorLog..

With these two changes the self-test on dev branch runs clean. The only odd thing is that the credentials test doesn't seem to work automatically anymore. I need to manually type bob/smith..

deploy/sample/triggers-config.sample.xml
```diff
@@ -0,0 +1,64 @@
...
+  <trgr:trigger>
+    <trgr:name>CreateTrigger</trgr:name>
+    <trgr:description>Trigger to enrich XML ingested via MLCP</trgr:description>
+    <trgr:data-event>
+      <trgr:directory-scope>
+        <trgr:uri>/myDir/</trgr:uri>
+        <trgr:depth>1</trgr:depth>
+      </trgr:directory-scope>
+      <trgr:document-content>
+        <trgr:update-kind>modify</trgr:update-kind>
```
grtjn: CreateTrigger but update-kind modify?

deploy/sample/triggers-config.sample.xml
```diff
@@ -0,0 +1,64 @@
...
+      </trgr:document-content>
+      <trgr:when>pre-commit</trgr:when>
+    </trgr:data-event>
+    <!--
+    <trgr:database-online-event>
+      <trgr:user-name>admin</trgr:user-name>
+    </trgr:database-online-event>
+    -->
```
grtjn: Superfluous, you also have the OnlineTrigger below..

deploy/sample/triggers-config.sample.xml
```diff
@@ -0,0 +1,64 @@
...
+      <trgr:database>${modules-db}</trgr:database>
+      <trgr:root>/triggers/</trgr:root>
+      <trgr:path>log-create.xqy</trgr:path>
+    </trgr:module>
+    <trgr:enabled>true</trgr:enabled>
+    <trgr:permissions>
+      <sec:permission>
+        <sec:capability>read</sec:capability>
+        <sec:role-name>${app-role}</sec:role-name>
+      </sec:permission>,
```
grtjn: Superfluous comma..

deploy/sample/triggers-config.sample.xml
```diff
@@ -0,0 +1,64 @@
...
+      <sec:permission>
+        <sec:capability>update</sec:capability>
+        <sec:role-name>${app-role}</sec:role-name>
+      </sec:permission>
+    </trgr:permissions>
+    <trgr:recursive>false</trgr:recursive>
+    <trgr:task-priority>normal</trgr:task-priority>
+  </trgr:trigger>
+  <trgr:trigger>
```
grtjn: How about some extra whitespace to make this XML easier to read?

deploy/sample/triggers-config.sample.xml
```diff
@@ -0,0 +1,64 @@
...
+    </trgr:data-event>
+    <!--
+    <trgr:database-online-event>
+      <trgr:user-name>admin</trgr:user-name>
+    </trgr:database-online-event>
+    -->
+    <trgr:module>
+      <trgr:database>${modules-db}</trgr:database>
+      <trgr:root>/triggers/</trgr:root>
+      <trgr:path>log-create.xqy</trgr:path>
```
grtjn: How about adding this trigger as sample code into the sample dir as well?

dmcassel:
Okay, I added triggers to self-test. Doesn't do a whole lot of validation at this point, but it does run it, so big errors should get caught.

deploy/test/test_server_config.rb
```diff
@@ -95,10 +95,16 @@
...
+#        sleep(10)
         @s.validate_install.must_equal true
       end
     end
 
+    it "should deploy triggers" do
+      @s.deploy_triggers.must_equal true
+      @s.clean_triggers.must_equal true
+    end
+
```
grtjn: Is there a triggers-config.xml, or is the self-test running with the sample one?

grtjn:
FYI, applied this for a project, deployment of triggers worked like a charm, thanks!

Though, could you get rid of this xdmp:log statement: https://github.com/dmcassel/roxy/blob/issue-109/deploy/lib/xquery/setup.xqy#L5507. It's debug info..

deploy/test/test_server_config.rb
dmcassel: This ran for me because I already had a triggers-config file in deploy. Just revised so that it used the one in the sample directory for the test.

dmcassel:
@grtjn have I addressed all your concerns?

grtjn:
I only notice one more thing:

You fixed the trigger config for ml7 self-test, but not for ml6, despite the fact you did enable triggers-db in ml6..

If you can attend to that, I'll try to run some elaborate self-tests tomorrow morning..

dmcassel:
Oops, missed ml6. Just added that.

grtjn:
Ran a self-test, but the deploy triggers is failing:

```
I, [2016-03-31T10:39:20.746472 #36838]  INFO -- : Deploying Triggers
E, [2016-03-31T10:39:20.752485 #36838] ERROR -- : ml.triggers.file=/Users/gjosten/Projects/github-grtjn/roxy/deploy/triggers-config.xml
E, [2016-03-31T10:39:20.752592 #36838] ERROR -- : Before you can deploy triggers, you must define a configuration. Steps:
1. Copy deploy/sample/triggers-config.sample.xml to /Users/gjosten/Projects/github-grtjn/roxy/deploy/triggers-config.xml
  The location of this file is controlled by the triggers.file property.
2. Edit /Users/gjosten/Projects/github-grtjn/roxy/deploy/triggers-config.xml to specify your trigger(s)
3. Run 'ml <env> deploy triggers')
```

Weird though since properties and all are looking correct. Need to take a closer look later..

grtjn:
We decided to move selftest changes out of this PR. It is ready to go..

Issue #581, opened 2016-03-11
How to set S3 credentials property throught Roxy


romain-nio:
Hi,

I'm using marklogic 8 and S3 for my scheduled backups.

I'm wondering if it's possible to configure AWS S3 crendentials (access key + secret key) directly through Roxy ?

I only found the way to configure this settings manually in the admin : 
og into the Admin Interface.
Click Security icon on the left tree menu.
Click Credentials to open the Credential Configuration page.
Enter the aws access key and aws secret key provided for your AWS account.

Thanks

grtjn:
Fixed in dev..

PR #582, opened 2016-03-11
Support bootstrapping and wiping AWS Credentials

RobertSzkutak:
Issue #581


dmcassel:
@RobertSzkutak I just ran a test with a non-AWS configuration and it failed. 

> ERROR: XDMP-AS: (err:XPTY0004) sec:credentials-set-aws($accessKey, $secretKey) -- Invalid coercion: fn:QName("","secretKey") as xs:string

The implementation of `setup:create-credentials` and `setup:wipe-credentials` needs to account for `$import-config/sec:credentials` not being present. If that's the case, the function should do nothing.

RobertSzkutak:
Thanks, should be fixed now!

dmcassel:
Looks good. I've tested that this works fine with non-AWS configuration. I don't have an AWS setup handy, but your code looks good to me.

PR #583, opened 2016-03-15
Fix #557

### LINKED ISSUE #557: Roxy ignores weight setting
On behalf of Laurent Tande:

There is a bug in roxy for the weight in fields : the namespace has been forgotten so weight is ignore and it’s always set to “1.0” when doing bootstrap :
https://github.com/marklogic/roxy/blob/dev/deploy/lib/xquery/setup.xqy#L1647

It should be change to
```
 admin:database-field-path($path/db:path,
 ($path/db:weight,
 1.0)[1]))
```

(Note the db: prefix in front of weight)

I tested it with this in ml-config.xml :
```
      <fields>
        <field>
          <field-name>Test</field-name>
          <field-path>
            <path>/some/xpath</path>
            <weight>4</weight>
          </field-path>
          <field-path>
            <path>/some/xpath2</path>
            <weight>2</weight>
          </field-path>
          <field-path>
            <path>/some/xpath3</path>
            <weight>0.5</weight>
          </field-path>
          <word-lexicons/>
          <included-elements/>
          <excluded-elements/>
          <tokenizer-overrides/>
          <stemmed-searches>basic</stemmed-searches>
          <fast-phrase-searches>true</fast-phrase-searches>
          <fast-case-sensitive-searches>true</fast-case-sensitive-searches>
          <fast-diacritic-sensitive-searches>true</fast-diacritic-sensitive-searches>
        </field>
      </fields>
```

RobertSzkutak:
#557 

Ran into this in the field today (haha, see the pun?) and the suggested fix by @grtjn did the trick.

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -1714,7 +1714,7 @@ declare function setup:add-fields-R(
                 $field/db:field-name,
                 for $path in $field/db:field-path
                 return
-                  admin:database-field-path($path/db:path, ($path/weight, 1.0)[1]))
+                  admin:database-field-path($path/db:path, ($path/db:weight, 1.0)[1]))
               )',
             (xs:QName("admin-config"), $admin-config,
              xs:QName("database"), $database,


dmcassel:
Looks good, thanks for sending

Issue #585, opened 2016-03-16
Support for schema, namespaces, module locations and traces for groups


rlouapre:
I do not believe Roxy deploy currently support these feature.
Please confirm.
I can provide a PR soon.

grtjn:
Correct, only on server level. You may be able to copy paste some bits from there..

grtjn:
Fixed in dev, thanks for the PR @rlouapre!

PR #586, opened 2016-03-17
issue 585

rlouapre:
Manually tested: module-location, namespace, schema on ML 8.0-4.2. #585

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -110,11 +110,11 @@ declare variable $group-settings :=
     <!--TODO: setting>audit-role-restriction</setting>
     <setting>audit-uri-restriction</setting>
     <setting>audit-user-restriction</setting-->
-    <setting function="add" keys="namespace-uri location">module-location</setting>
-    <setting function="add" keys="prefix namespace-uri">namespace</setting>
-    <setting function="add" keys="namespace-uri schema-location">schema</setting>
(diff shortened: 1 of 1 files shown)


dmcassel:
@grtjn I think you've done the most with group settings; how does this look to you?

grtjn:
Looks like smart code to me. Interesting thought to leverage new attributes on the setting elements. Definitely need to run self-test with this (e.g. I think we need to extend ml7-config.xml), and I'd like to take a closer look at the xdmp:value logic as well..

rlouapre:
I am happy to provide some tests but I have done it long time ago and I cannot remember how to run them and what to implement.  
Do we have documentation on how to write and run the tests?

grtjn:
It is the self-test command (`./ml self-test`). It by default runs with the deploy/test/data/ml7-config.xml (which we also use to test ml8 at the moment) against localhost. You can put a local.properties in deploy/test/data/ml7-properties/ to override that if necessary..

grtjn:
Not aware of documentation on self-test unfortunately..

rlouapre:
Some fixes and new validation used by ```self-test```. See new PR #590

grtjn:
Funny, you normally don't need a new PR to append commits to it..

rlouapre:
Agree. Not sure what happens...

grtjn:
Closing this in favor of #590..

Issue #416, opened 2015-04-21
/roxy/rewrite.xqy broken with ML 8.0-2


aajacobs:
Roxy's /roxy/rewrite.xqy relies on a conf:rewrite function in ML server that was removed in the ML8.0-2 release.  As a result, any app using the /roxy/rewrite.xqy will begin showing error similar to the one below:
```
<error:error xsi:schemaLocation="http://marklogic.com/xdmp/error error.xsd" xmlns:error="http://marklogic.com/xdmp/error"
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <error:code>XDMP-UNDFUN</error:code>
  <error:name>err:XPST0017</error:name>
  <error:xquery-version>1.0-ml</error:xquery-version>
  <error:message>Undefined function</error:message>
  <error:format-string>XDMP-UNDFUN: (err:XPST0017) Undefined function conf:rewrite()</error:format-string>
  <error:retryable>false</error:retryable>
  <error:expr> </error:expr>
  <error:data>
    <error:datum>conf:rewrite()</error:datum>
  </error:data>
  <error:stack>
    <error:frame>
      <error:line>7</error:line>
      <error:column>9</error:column>
      <error:operation>xdmp:eval("&amp;#10;        import module namespace conf = &amp;quot;http://marklog...", (fn:QName("","method"), "GET", fn:QName("","uri"), ...))</error:operation>
      <error:xquery-version>1.0-ml</error:xquery-version>
    </error:frame>
    <error:frame>
      <error:uri>/roxy/rewrite.xqy</error:uri>
      <error:line>40</error:line>
      <error:column>6</error:column>
      <error:variables>
	<error:variable>
	  <error:name xmlns="">uri</error:name>
	  <error:value>"/web/scripts/d3js/crossfilter.v1.min.js"</error:value>
	</error:variable>
	<error:variable>
	  <error:name xmlns="">method</error:name>
	  <error:value>"GET"</error:value>
	</error:variable>
	<error:variable>
	  <error:name xmlns="">path</error:name>
	  <error:value>"/web/scripts/d3js/crossfilter.v1.min.js"</error:value>
	</error:variable>
	<error:variable>
	  <error:name xmlns="">final-uri</error:name>
	  <error:value>()</error:value>
	</error:variable>
      </error:variables>
      <error:xquery-version>1.0-ml</error:xquery-version>
    </error:frame>
  </error:stack>
</error:error>
```

[...]

PR #587, opened 2016-03-17
Issue 416

dmcassel:
I took a similar approach to what other people had posted. Because the rule functions have changed over time, I dynamically loaded the conf:*-rule() functions. Rewriting needs to be fast, so to avoid doing that on every call, I put the results into a server field. 

@tdiepenbrock, @grtjn, @joeg13, @RobertSzkutak, @garyrusso, @aajacobs -- anybody available to test in different scenarios? 

I encountered #416 trying to run unit tests with ML 8.0-4 with an MVC project.


dmcassel:
Accidentally submitted against the master branch.

PR #588, opened 2016-03-17
#416 proposed solution with dynamic rule lookup

dmcassel:
I took a similar approach to what other people had posted. Because the rule functions have changed over time, I dynamically loaded the conf:*-rule() functions. Rewriting needs to be fast, so to avoid doing that on every call, I put the results into a server field.

@tdiepenbrock, @grtjn, @joeg13, @RobertSzkutak, @garyrusso, @aajacobs -- anybody available to test in different scenarios?

I encountered #416 trying to run unit tests with ML 8.0-4 with an MVC project.

### DIFF
--- a/src/roxy/lib/rewriter-lib.xqy
+++ b/src/roxy/lib/rewriter-lib.xqy
@@ -1,0 +1,196 @@
+xquery version "1.0-ml";
+
+module namespace rewriter = "http://marklogic.com/roxy/rewriter";
+
+import module namespace conf = "http://marklogic.com/rest-api/endpoints/config"
+  at "/MarkLogic/rest-api/endpoints/config.xqy";
+import module namespace rest = "http://marklogic.com/appservices/rest"
+  at "/MarkLogic/appservices/utils/rest.xqy";
+import module namespace eput = "http://marklogic.com/rest-api/lib/endpoint-util"
+  at "/MarkLogic/rest-api/lib/endpoint-util.xqy";
+
+declare variable $RULE-SERVER-FIELD := "roxy:rewrite-rules";
+
+(: This is a hack, but allows us to maintain one rule set endpoints and rewriter,
+    while pushing the finer-grained error handling to the endpoint level. :)
+declare function rewriter:rewrite-rules(
+) as map:map
+{
+  let $old-rules := eput:get-rest-options()
+  return
+    if (exists($old-rules))
+    then $old-rules
+    else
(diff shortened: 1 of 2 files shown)


grtjn:
Haven't fully wrapped my mind around this yet, but a question did pop up: this assumes the presence of conf:*-rules() functions. Isn't it likely they will disappear in future as well, since everything is now recorded in rewriter.xml? Would it make sense to somehow consume that file instead? Or maybe generate a new customized rewriter.xml that appends the Roxy routes?

RobertSzkutak:
I've actually played around with trying to get rewriter.xml working to route to Roxy but it's not trivial. Even though I've gotten it 75% working the problem is that you end up in an "infinite loop" in a lot of situations where rewriter.xml reroutes to rewriter.xqy who's routed request then hits rewriter.xml etc. You also have problems pulling the originally requested URI and parameters because the incoming request to rewriter.xqy comes from rewriter.xml.

After implementing a lot of tricks (passing the original route as a param from rewriter.xml, modifications to the rewriter.xqy, careful regex, etc.) I got most of the routes working but not all. 

I could push this into a new branch on my fork if you're interested in taking a look.

Ultimately, after working away on adding rewriter.xml, I think the best thing would be to make a substantial change to routing so that it's handled exclusively by rewriter.xml or rewriter.xqy. @dmcassel has done the latter and I fully support it for an immediate solution. Long-term, as @grtjn points out, it's probably worth trying to migrate to rewriter.xml to avoid this happening again.

grtjn:
Any attempts should interpret rewriter.xml instead of making rewriter.xml point to roxy rewriter or vice versa. One of both needs to take control, or to put it more precisely: one of both needs to take full control, because you can no longer invoke one from the other in a simple elegant way..

RobertSzkutak:
I gave this solution to another consultant who in turn gave it to a large health insurer today (not the one y'all probably affiliate me with) to help them migrate a hybrid app from ML7 to ML8. Worked like a charm in all of their testing.

dmcassel:
@RobertSzkutak excellent!

@grtjn @paxtonhare what do you think about accepting this PR?

grtjn:
I still need a spare moment to look closer, just back from some time away..

grtjn:
Would also solve #599. The comment from Robert sounds very convincing. I wouldn't mind if we merge this. It only touches mvc and hybrid anyhow, which are used less these days..

grtjn:
@garyrusso confirmed in #599. Merging..

Issue #593, opened 2016-04-04
deploy_rest doesn't use REST API to deploy options


dmcassel:
REST API options ought to be deployed by [PUTting them to /v1/config/query/{name}](http://docs.marklogic.com/REST/PUT/v1/config/query/['default'-or-name]), but instead Roxy is manually constructing the URIs and writing them to the modules database. That bypasses the validation that happens when loading the options and puts us at risk for the diverging with how the REST API works. (The divergence problem hasn't happened yet, but improperly bypassing validation has.)

dmcassel:
fixed in dev

PR #597, opened 2016-04-07
#593 Change deployment of REST options to use REST API

dmcassel:
Implementation now uses REST API, so options will get validated

### DIFF
--- a/deploy/lib/ml_rest.rb
+++ b/deploy/lib/ml_rest.rb
@@ -95,7 +95,7 @@ module Roxy
 
           contents = file.read
 
-          r = go("http://#{@hostname}:#{@port}/v1/config/query/#{basename}", "put", headers, nil, contents)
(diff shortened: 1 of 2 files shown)


deploy/lib/ml_rest.rb
```diff
@@ -73,6 +73,37 @@ def install_properties(path)
...
+            headers['Content-Type'] = 'application/xml'
+          elsif (ext == '.json')
+            headers['Content-Type'] = 'application/json'
+          else
+            @logger.error("Unrecognized REST options format: #{item}")
+          end
+
+          contents = file.read
+
+          r = go("http://#{@hostname}:#{@port}/v1/config/query/#{basename}", "put", headers, nil, contents)
```
paxtonhare: Look at how we are doing https or http in server_config.rb. You should probably do that here.

dmcassel: Fixed here and fixed other references in this file

Issue #438, opened 2015-05-11
Use different credentials for ml {action} modules


rlouapre:
Is there a way to provide different credentials for ```ml {action} modules```?

I currently need to create Roxy another environment and override ```user``` and ```password``` properties from the command line.

It could even make sense to only use ```admin``` for ```ml bootstrap```

grtjn:
Interesting. How about bootstrap-user/-password, deploy-user/-password, both defaulting to user/password. First to be used for bootstrap/wipe/restart, the other for anything else (deploy/clean/app_specific)?

rlouapre:
:+1:

grtjn:
Nice to have, pushing away for the moment.. (should be easy though..)

grtjn:
Whoops, accidentally closed this..

RobertSzkutak:
PR For this now merged into dev. Separate users now supported for bootstrap, deploy, mlcp and corb.

PR #602, opened 2016-04-24
Support different user credentials for different Roxy operations - #438

RobertSzkutak:
Implementation of the proposal to support #438 and somewhat related to #624

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -631,6 +631,9 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
   def bootstrap
     raise ExitException.new("Bootstrap requires the target environment's hostname to be defined") unless @hostname.present?
 
+    @ml_username = @properties['ml.bootstrap-user'] || @properties['ml.user']
+    @ml_password = @properties['ml.bootstrap-password'] || @properties['ml.password']
+
     internals = find_arg(['--replicate-internals'])
     if internals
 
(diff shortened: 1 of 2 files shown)


grtjn:
Odd, github shows a lot of diffs, must be whitespace/lineends? Use this to see the real diff:

https://github.com/marklogic/roxy/pull/602/files?w=1

Looks clean to me at first glance, but haven't tried yet..

RobertSzkutak:
Thanks! I swear up and down my sublime settings match (or at least, should match) the recommendations provided. I have no idea why those diffs are in there or how I can clear them out. Open to any and all suggestions.

deploy/lib/server_config.rb
```diff
@@ -728,6 +734,10 @@ def bootstrap
   end
 
   def wipe
+
+    @ml_username = @properties['ml.bootstrap-user'] || @properties['ml.user']
+    @ml_password = @properties['ml.bootstrap-password'] || @properties['ml.password']
```
dmcassel: I think password selection needs to be a little more involved. 

    if ml.bootstrap-user, then use ml.bootstrap-password
      use it even if it's empty, which will result in calling password_prompt
    if !ml.bootstrap-user, then fall back to ml.password

Likewise for deploy-password below. This will handle the situation where some other password is used with the bootstrap user (unlikely, but quite possible).

RobertSzkutak: Great catch! I will make these changes and update the PR soon.

RobertSzkutak:
@dmcassel I confirmed and then fixed the edge case you found. I also added support for a corb user and mlcp user properly. Look good to merge?

dmcassel:
Looks good!

Issue #601, opened 2016-04-22
Space in Ruby Path causes ml.bat to fail


rjkennedy98:
After doing Roxy upgrade to 1.7.3 master the ml.bat script fails in windows.

C:\EDH_ML_Path4_3\eie_edh_framework\EDH_ML>C:\Program Files (x86)\Ruby21\bin\rub
y.exe -Ideploy -Ideploy\lib deploy\lib\ml.rb
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

dmcassel:
I don't have a Windows setup handy. @rjkennedy98 could you run a quick test? In [ml.bat, line 12](https://github.com/marklogic/roxy/blob/master/ml.bat#L12), add quotes around the value for RUBYFOUND, so that it looks like this:

    set RUBYFOUND="%%~$PATH:X"

If that works, we can update the script. If not, I invite input from someone with more Windows scripting fu than I have.

rjkennedy98:
That gave me:
C:\EDH_ML_Path4_3\eie_edh_framework\EDH_ML>ml local deploy modules
'""' is not recognized as an internal or external command,

I have made a change ml.bat, line 158
  "%RUBYFOUND%" -Ideploy -Ideploy\lib -Ideploy\test deploy\test\test_main.rb

and ml.bat, line 167
 "%RUBYFOUND%" -Ideploy -Ideploy\lib deploy\lib\ml.rb %*

that seemed to resolve issue

rlouapre:
I am not sure we should have provided a fix for it...
Screen from Ruby installer:
![image](https://cloud.githubusercontent.com/assets/6400566/15516773/5f9d7abc-21c2-11e6-88e4-8a826990ae67.png)
It clearly says "please avoid any folder name that contains spaces"

dmcassel:
@rlouapre interesting find. While I agree with you -- the instructions are pretty clear -- it's a pretty simple fix for Roxy. Might as well.

rjkennedy98:
The client I am working with has managed VMs which have a space in the ruby install path. Otherwise, yes, it could be easily avoided by following directions.

grtjn:
Fixed in dev..

PR #607, opened 2016-04-28
#601 adding quotes around the ruby path

dmcassel:
#601

### DIFF
--- a/ml.bat
+++ b/ml.bat
@@ -155,7 +155,7 @@ goto end
     goto :loop2
   )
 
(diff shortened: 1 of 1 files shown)


grtjn:
Looks good. Would have to grab hold of my win laptop to test this..

dmcassel:
Can someone test this on Windows?

rlouapre:
That seems to fix the issue - but please also note my comment. There are best practices on the installing Ruby on Windows...

dmcassel:
@rlouapre thanks for the test. While acknowledging the best practice, it looks like this is still a useful change.

Issue #612, opened 2016-05-17
Move Roxy off Query Console endpoint


clockworked247:
Roxy should be using the /v1/eval endpoint instead.  ML 9 will add deeper security to QC endpoints, no longer allowing for Roxy or any other unauthorized requestors to use them.

dmcassel:
+1 on this. We can probably do it for ML8 as well, since that's when the endpoint became available.

grtjn:
Fixed in dev..

Issue #596, opened 2016-04-06
Test Roxy deployment with MarkLogic 9


dmcassel:
MarkLogic 9 Early Access 1 release is out (limited availability). Test and fix as needed.

grtjn:
There seems to be (yet another) small change in QC evaler. Apparently, it now requires both dbid and sid. I changed execute_query_7 in deploy/lib/server_config.rb as follows, and that seems to run against ML9. Not entirely convinced it is backwards-compatible:

    def execute_query_7(query, properties = {})
      # We need a context for this query. Here's what we look for, in order of preference:
      # 1. A caller-specified database
      # 2. A caller-specified application server
      # 3. An application server that is present by default
      # 4. Any database
      if properties[:db_name] != nil
        db_id = get_db_id(properties[:db_name])
    +   sid = get_sid("Manage")
      elsif properties[:app_name] != nil
        sid = get_sid(properties[:app_name])
    +   db_id = get_db_id("Documents")
      else
        sid = get_sid("Manage")
    +   db_id = get_db_id("Documents")
      end

      ws_id, q_id = create_roxy_workspace()
      raise ExitException.new("Can't create Roxy workspace in QConsole") unless ws_id && q_id

      db_id = get_any_db_id if db_id.nil? && sid.nil?

      # necessary to work around weirdness on windows
      headers = {
        'content-type' => 'text/plain'
      }

      url = "#{@protocol}://#{@hostname}:#{@qconsole_port}/qconsole/endpoints/evaler.xqy?wsid=#{ws_id}&qid=#{q_id}&action=eval&querytype=xquery&dirty=true"
      if db_id.present?
        url = url + "&dbid=#{db_id}"
    + end
    + if sid.present?
        url = url + "&sid=#{sid}"
      end
    + r = go(url, "post", headers, nil, query)

      delete_workspace(ws_id) if ws_id

      raise ExitException.new(JSON.pretty_generate(JSON.parse(r.body))) if r.body.match(/\{"error"/)

      r
    end

dmcassel:
I'm working on a solution to this using /v1/eval.

dmccreary:
Thanks for the fix.  Just got this working on a new ML-9 test project.

grtjn:
Dev branch now contains a better way of executing queries against ML8 and 9. Closing this..

PR #613, opened 2016-05-19
Issue 612: user /v1/eval for ML8+

dmcassel:
Addresses #612 and #596. Uses /v1/eval for MarkLogic 8 and MarkLogic 9.

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -506,8 +506,10 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
       r = execute_query_4 query, properties
     elsif @server_version == 5 || @server_version == 6
       r = execute_query_5 query, properties
-    else # 7 or 8
+    elsif @server_version == 7
       r = execute_query_7 query, properties
+    else # 8 or 9
+      r = execute_query_8 query, properties
     end
 
     raise ExitException.new(r.body) unless r.code.to_i == 200
@@ -2114,6 +2116,34 @@ private
     end
 
     delete_workspace(ws_id) if ws_id
+
+    raise ExitException.new(JSON.pretty_generate(JSON.parse(r.body))) if r.body.match(/\{"error"/)
+
+    r
+  end
+
+  def execute_query_8(query, properties = {})
+    # headers = {}
+
+    # params = {
+    #   :xquery => query,
+    #   :locale => LOCALE,
+    #   :tzoffset => "-18000"
+    # }
+
+    # r = go "#{@protocol}://#{@hostname}:#{@qconsole_port}/eval", "post", headers, params
+
+    headers = {
+      "Content-Type" => "application/x-www-form-urlencoded"
+    }
+
+    params = {
+      :xquery => query,
+      :locale => LOCALE,
+      :tzoffset => "-18000"
+    }
+
+    r = go "#{@protocol}://#{@hostname}:#{@qconsole_port}/v1/eval", "post", headers, params
 
     raise ExitException.new(JSON.pretty_generate(JSON.parse(r.body))) if r.body.match(/\{"error"/)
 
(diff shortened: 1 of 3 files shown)


deploy/lib/server_config.rb
```diff
@@ -2118,6 +2120,34 @@ def execute_query_7(query, properties = {})
...
+
+    params = {
+      :xquery => query,
+      :locale => LOCALE,
+      :tzoffset => "-18000"
+    }
+
+    r = go "#{@protocol}://#{@hostname}:#{@qconsole_port}/v1/eval", "post", headers, params
+
+    raise ExitException.new(JSON.pretty_generate(JSON.parse(r.body))) if r.body.match(/\{"error"/)
```
paxtonhare: Did you verify that this line works as intended? I'm ignorant of the output of /v1/eval.

grtjn:
I'm amazed, is that all? I'm kinda missing context like db or app-server selection. We have a lot of code that picks targets explicitly. Does that still work?

I guess we could run self-test to see what comes out of that, but also thinking of app_specific, in which exec of queries is used a lot..

grtjn:
`/v1/eval` returns `multipart/mixed` response. We need to parse that similarly to how we were parsing the json response from the newer QC. I ran a test and this code seemed to work nicely, which I added to `util.rq`:

```ruby
def parse_multipart(body)
  if (body.match("^\r\n--"))
    # Extract the delimiter from the response.
    delimiter = body.split("\r\n")[1].strip
    parts = body.split(delimiter)

    # The first part will always be an empty string. Just remove it.
    parts.shift
    # The last part will be the "--". Just remove it.
    parts.pop

    # Get rid of part headers
    parts = parts.map{ |part| part.split("\r\n\r\n")[1].strip }

    # Return all parts as one long string, like we were used to.
    return parts.join("\n")
  else
    return body
  end
end

def parse_body(body)
  parse_multipart(parse_json(body))
end
```

And then replace references to `parse_json` with `parse_body` in `server_config.rb` (dozen times or more), and `test_server_config.rb` (only once). Note: not in `editor.rb`, that has its own `parse_json`. I think that has a different purpose..

This should be enough to at least get self-test working. Worked for me locally.

grtjn:
Still want to look a bit closer to that context stuff. Stay tuned..

grtjn:
I searched through the Roxy code, and the app_name parameter is not used internally. db_name is though, so we need to leverage the database param of /v1/eval..

grtjn:
I filed an RFE to request server context support for /v1/eval, I suggest we add the following to the top of `execute_query_8`:

```ruby
    if properties[:app_name] != nil
      raise ExitException.new("Executing queries with an app_name (currently) not supported with ML8+")
    end
```

And for database support we only need this between `params = ..` and `r = go ..`:

```ruby
    if properties[:db_name] != nil
      params[:database] = properties[:db_name]
    end
```

grtjn:
@dmcassel Did you incorporate my suggestions yet?

dmcassel:
Okay, I made @grtjn's suggested updates. self-test on ml7 ran successfully, passed my testing on ml8 and ml9.

grtjn:
Commits look clean, I'll try to give it a spin asap..

grtjn:
Yep, self-test runs clean at my end too..

Just wondering, we may wanna apply parse_body to the result of eval itself as well, otherwise this might fail?

```
raise ExitException.new(JSON.pretty_generate(JSON.parse(r.body))) if r.body.match(/\{"error"/)
```

(this line exists in execute_query_5, execute_query_7, and execute_query_8)

dmcassel:
@grtjn I set up bootstrap to fail a couple different ways, but saw no change when I modified or even removed that line from execute_query_8. The errors were displayed correctly. Not sure what kind of error gets to that line, but it doesn't seem to need the change.

grtjn:
good enough for me..

grtjn:
let's wait no longer..

dmcassel:
Thanks @grtjn!

PR #614, opened 2016-05-24
Allow amps to be updated and not just created

tapuzzo-fsi:
Previously, existing amp definitions were not changed by bootstrap, now the roles are updated when bootstrapping.

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -4761,8 +4761,7 @@ declare function setup:create-amps($import-config)
         <options xmlns="xdmp:eval">
           <database>{xdmp:security-database()}</database>
         </options>
-      ),
-      setup:add-rollback("amps", $amp)
+      )
     )
     else
     (
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -4744,7 +4744,26 @@ declare function setup:create-amps($import-config)
     if ($existing-amps/sec:amp[sec:namespace = $amp/sec:namespace and
                                    sec:local-name = $amp/sec:local-name and
                                    sec:document-uri = $amp/(sec:doc-uri, sec:document-uri) and
-                                   sec:db-name = $amp/sec:db-name]) then ()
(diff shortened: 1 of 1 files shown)


deploy/lib/xquery/setup.xqy
```diff
@@ -4744,7 +4744,26 @@ declare function setup:create-amps($import-config)
...
+           $amp/(sec:doc-uri, sec:document-uri)[1],
+           $db,
+           $amp/sec:role-name
+        )',
+        (xs:QName("amp"), $amp),
+        <options xmlns="xdmp:eval">
+          <database>{xdmp:security-database()}</database>
+        </options>
+      ),
+      setup:add-rollback("amps", $amp)
```
dmcassel: An error condition will result in the amp getting wiped out during rollback, rather than restored to its original roles. I think it's better to not call setup:add-rollback() here, analogous to [setup:apply-database-settings()](https://github.com/marklogic/roxy/blob/dev/deploy/lib/xquery/setup.xqy#L1507).

TonyApuzzo: OK, that makes sense, I wasn't sure if add-rollback was a removal or restore function.  I'll remove that line.  Thanks for the review.

tapuzzo-fsi:
OK, I updated the PR based on that suggestion. Totally makes sense, thanks for the review.

grtjn:
Hmm, just wondering: can you specify multiple roles for an amp? We usually tend to add a wrapping element if multiple would be allowed. I think it would work, but just pondering we lack the wrapper. Makes me think whether the built-in format would use that or not, as that is probably what would be returned by capture (though not sure capture returns amps)..

dmcassel:
[sec:amp-set-roles()](http://docs.marklogic.com/sec:amp-set-roles) does take `$role-names as xs:string*`. The submitted code is consistent with how amps are created, so I think we're okay there.

dmcassel:
@tapuzzo-fsi thanks for the PR!

tapuzzo-fsi:
Yes, I've tested this with multiple roles and it works as expected.  Thanks for accepting the PR.

PR #617, opened 2016-06-08
Fix test-helper assert-throws-error().

tapuzzo-fsi:
The original fix for #503 (#526) didn't work for me and all the unit tests using `assert-throws-error()` fail on my installation of ML 8.0.1-1.  This PR change works for my version of ML and should work on newer versions since it directly calls the wrapped helper rather than building a cascade of empty-sequence calls.

There is a quirk in json:to-array() in that it collapses the input into a single-dimension array when its arguments are a one element array.  This PR works around that quirk and allows a single virtualized vararg to be sent correctly.

### DIFF
--- a/src/test/test-helper.xqy
+++ b/src/test/test-helper.xqy
@@ -278,37 +278,37 @@ declare function helper:assert-meets-maximum-threshold($expected as xs:decimal, 
 
 declare function helper:assert-throws-error($function as xdmp:function)
 {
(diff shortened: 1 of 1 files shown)


grtjn:
Pity the cascade didn't work out as well as hoped, but thanks for testing and providing a fix. I vote to merge this, hardly touches anyone anyhow..

dmcassel:
Looks fine to me.

Issue #521, opened 2015-09-21
support custom properties for corb2


joemfb:
It'd be nice if we could pass custom properties to corb: https://stackoverflow.com/questions/31751243/marklogic-8-corb-pass-arguments-other-than-uri

Also, we should probably be bundling/supporting corb2,  if we're not already ...

grtjn:
If I read and remember correctly, corb doesn't support such properties, so this means we would have to implement support for corb2 then. Makes sense..

dmcassel:
@hansenmc with your pull request, this issue can be closed, right?

hansenmc:
Definitely if the custom properties are set in an options file, but I think I may need to make a slight adjustment to ensure that if passed in via commandline, a lower or camelCased property does not get turned into UPPER-CASE (e.g. URIS-MODULE.foo => URIS-MODULE.FOO). I can get that added this evening.

grtjn:
Above mentioned fix was added and merged, and latest corb jar with usage included as well!

PR #626, opened 2016-07-14
Fixing Issue #521: support custom properties for CoRB2

hansenmc:
Updated the corb task to use CoRB2 v2.3.1 and support all of the CoRB2 options.

Ensured that the current Roxy commandline options and functionality still work,
and also support specifying any of the CoRB2 options with either "--" or "-D"
prefix, case-insensitive (i.e. --options-file or -DOPTIONS-FILE).

Removed the explicit checks/errors for Roxy corb task required fields, since there
are now additional options available: URIS-FILE, XML-FILE, PROCESS-MODULE (now 
preferred instead of XQUERY-MODULE), specifying all options inside of 
OPTIONS-FILE, etc.

Also upgraded the XCC jar to v8.0.5

#521


deploy/lib/server_config.rb
```diff
@@ -1142,38 +1142,150 @@ def corb
...
+    if !module_root.blank?
+        options["MODULE-ROOT"] = module_root
+    end
+
+    modules_database = @properties['ml.modules-db'] || ''
+    if !modules_database.blank?
+        options["MODULES-DATABASE"] = modules_database
+    end
+
+    # A full list of CoRB2 options
```
grtjn: Just wondering, instead of adding knowledge of all Corb2 options inside Roxy, isn't is easier to just pass through whatever gets specified, and let Corb squeak itself if something is wrong? A bit like how we did things for MLCP, in which we more or less only append user/pwd/etc, and parse options files for properties replacement.

grtjn:
Thanks for the PR! Wondering though if we could do with less lines. See inline comment..

grtjn:
Looking great at first glance!

deploy/lib/server_config.rb
```diff
@@ -1142,38 +1142,77 @@ def corb
...
+                        .join(' ')
 
     # Find the jars
-    corb_file = find_jar("corb")
-    xcc_file = find_jar("xcc")
+    corb_file = find_jar('corb')
+    xcc_file = find_jar('xcc')
+
+    runme = %Q{java -cp #{corb_file}#{path_separator}#{xcc_file} #{systemProperties} com.marklogic.developer.corb.Manager #{connection_string}}
+    logger.info runme
```
grtjn: Can you change this into logger.debug? You then only see that when you use -v..

deploy/lib/Help.rb
```diff
@@ -398,30 +398,27 @@ def self.corb
...
+      CoRB2 supports options files natively using the --OPTIONS-FILE parameter.
 
-      Corb Options:
-        --threads=1                     # the thread count to use
-        --root=/                        # the root of the modules database
-        --install=false                 # whether or not to install (default: false)
+      See: https://github.com/marklogic/corb2
 
       General options:
-        -v, [--verbose]  # Verbose output
```
grtjn: I think mentioning -v could be useful, particularly with the logger.debug below..

grtjn:
Just two small suggestions, and I think this is good to go..

grtjn:
PS: I created a ticket against Corb2 to improve usage printing to console. I like the way MLCP does it very much..

grtjn:
I'd be happy to merge this now. Ran a superficial test, and invoking Corb2 (to get a message on screen) succeeded. Low impact too. @paxtonhare / @dmcassel agree?

I'm only wondering if it is worth holding of for https://github.com/marklogic/corb2/issues/31 to be fixed, so we can integrate a slightly newer Corb2 jar with better usage console output..

dmcassel:
I think this looks good (haven't tested). Is this change backwards compatible?

dmcassel:
Regarding marklogic/corb2#31, I don't see any comments there suggesting when it might get done. I agree it would be helpful, but I would suggest not waiting on it unless a short timeline gets proposed.

hansenmc:
Corb2 is mostly backwards compatible, except the built-in uris module is no longer provided. I tried to make it backwards compatible by continuing to support roxy corb params that did not already get mapped or have the same defaults, and automatically added an inline query when collection param was used.

I'll look at improving the corb2 exception. We could try toget it in and out out a maintenance release, if you need/want. Otherwise, we don't have a release planned in the immediate timeframe.

dmcassel:
Sounds good to me. @grtjn you ran some tests; care to do the honors?

grtjn:
Merging now.

@hansenmc I leave it up to you to do a PR with a newer Corb2 when it comes out. Fixing the console output would help end users a lot though, so I'd suggest not waiting long with that..

PR #635, opened 2016-07-22
Issue521 adjusted handling of commandline options to preserve case of custom input

hansenmc:
I adjusted the regex to add an additional match group and ensure that .upcase is only applied to the base option name and not the custom input. This will ensure that custom inputs are applied properly when specified on the commandline.


grtjn:
I think this could use a git rebase, bit hard to distinguish what is the real change here..

deploy/lib/server_config.rb
```diff
@@ -1182,11 +1182,11 @@ def corb
     end
 
     # collect options with either "--" or "-D" prefix, and normalize options to be UPPER-CASE
-    optionArgPattern = /^(--|-D)(.*?)="?(.*)"?/
+    optionArgPattern = /^(--|-D)(([^.]*?)(\..*?)?)="?(.*)"?/
```
grtjn: You have nested groups in here, which you probably don't need..

hansenmc: Agreed, I'll remove the wrapping group

grtjn:
Hmm, still too many commits in this PR. I think that is because this PR is against master branch instead of dev.

@hansenmc Could you open a new PR with your changed, but against dev branch?

hansenmc:
Whoops. I'll take a look tonight and get it cleaned up. 

Also, I did commit changes for better usage statements in CoRB2 to our development branch. There are a handful of little enhancements and bug fixes that are ready to ship, so I'll probably put out a point release and have this reference the new version, unless @bbandlamudi wants to hold off.

bbandlamudi:
Hi Mads, if we have enough changes for 2.3.2, we can release it. We need to think of new additions and I would like to hash them out with you on email.

PR #644, opened 2016-08-19
Issue521

hansenmc:
I adjusted the regex to add an additional match group and ensure that `.upcase` is only applied to the base option name and not the custom input. This will ensure that custom inputs are applied properly when specified on the commandline.

Also, updated to the newly released CoRB2 v2.3.2 with improved usage messages.


grtjn:
Thanks, looks pretty clean. Also thanks for upgrading the corb jar. I'll try to run a quick test later today, looking forward to it..

grtjn:
The new jar gives an elaborate usage when running `./ml local corb`, very helpful, excellent!

Not sure how to test the other change, but it is such minimal, and it looked correct, so merging!

Issue #642, opened 2016-08-16
xdbc-authentication-method should default to 'digest'


grtjn:
That used to be the hard-coded default, but in one of the later versions we made it configurable, and now the default depends on authentication-method. If that gets set to application-level, bootstrap will fail for the xdbc-server. Adding xdbc-authentication-method=digest to defaults.properties will prevent this..

grtjn:
Kind of duplicate of #625..

PR #646, opened 2016-08-23
Issue #642: making xdbc-authentication-method default to 'digest'

RobertSzkutak:
#642

### DIFF
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -194,6 +194,10 @@ use-https=false
 authentication-method=digest
 default-user=${app-name}-user
 #
+# The authentication used for your XDBC server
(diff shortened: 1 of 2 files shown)


grtjn:
Looks simple enough. :)

Thinking of a simple test. I guess generating an mvc project (since rest doesn't come with src), and bootstrapping with app-level auth + a deploy modules should be enough..

grtjn:
Works. Created an mvc project configured for app-level auth. Make sure to not use install-xcc=false. Bootstrap first failed, but after installing this branch with ./ml upgrade --fork=RobertSzkutak --branch=rob, it ran just fine..

Issue #620, opened 2016-06-20
ERROR: no implicit conversion of nil into String


grtjn:
Running `ml dev info` when `deploy/dev.properties` does not exist results in `ERROR: no implicit conversion of nil into String`. Line 131 of `server_config.rb` should read:

```
        logger.info k + ": " + v.to_s
```

(e.g. with an extra `.to_s` after v)

grtjn:
Note: `ml dev info --format=json` works fine..

grtjn:
PS: my impression was that it breaks on `ml.server` being not properly set..

RobertSzkutak:
Looks like a duplicate of issue #224

dmcassel:
fixed in dev

Issue #224, opened 2014-04-29
Running ml info on non-existing env breaks


grtjn:
Running `ml info prod` when you don't have prod.properties returns:

```
ERROR: no implicit conversion of nil into String
ERROR: ["deploy/lib/server_config.rb:101:in `+'", "deploy/lib/server_config.rb:101:in `block in info'", "deploy/lib/server_config.rb:100:in `each'", "deploy/lib/server_config.rb:100:in `info'", "deploy/lib/ml.rb:128:in `<main>'"]
```

dmcassel:
fixed in dev

PR #647, opened 2016-08-23
Fix for #620 and #224: prevent implicit conversion of nil to string

RobertSzkutak:
I implemented two fixes:

1) Properly setting ml.server
2) Changing v to v.to_s

In my testing, either fix appeared to independently solve #620 and #224

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -65,7 +65,9 @@ class ServerConfig < MLClient
     @config_file = @properties["ml.config.file"]
 
     @properties["ml.server"] = @properties["ml.#{@environment}-server"] unless @properties["ml.server"]
-    if (@properties["ml.server"] == nil) then @properties["ml.server"] = "" end
+    if (@properties["ml.server"] == nil)
+      raise "Error! ml.server not set. You may be missing deploy/" + @properties["environment"] + ".properties"
+    end
 
     @hostname = @properties["ml.server"]
     @bootstrap_port_four = @properties["ml.bootstrap-port-four"]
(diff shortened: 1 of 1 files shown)


grtjn:
I am actually thinking, why not just throw an error if ml.server isn't properly configured? Not sure if that will have side-effects on commands that don't need an env, but if it doesn't it would give proper feedback about properties being mal-configured..

grtjn:
I'd still keep the second change by the way, just change the first into an exception..

RobertSzkutak:
Pushed in a commit with your suggestion @grtjn

grtjn:
Don't always blindly believe my suggestions.. ;-)

We should run some tests with this. I recall some feature in latest slush-marklogic-node depends on ./ml dev info returning results without dev.properties. Also not sure what happens with commands that don't take env (like ml upgrade and such)..

RobertSzkutak:
I tried out every command and couldn't find any negative impact to upgrade, the general commands or the scaffolding commands. The bootstrapping commands and deployment/data commands app all appeared to be positively impacted. I learned that the settings command and some others threw nasty exceptions that were hard to parse as missing *.properties files.

I will investigate the impact on Slush soon.

dmcassel:
Looks good to me. @grtjn you happy with this?

RobertSzkutak:
@grtjn I did a review of Slush and can't find any code to suggest it would be impacted by this. I tried it out on a Slush install I already had up and running and it seemed ok. If Slush is impacted by this for some reason I'd personally prefer it finds another way to accomplish whatever it needs than by relying on returning non-existent properties..

dmcassel:
This looks good to me.

Issue #648, opened 2016-08-23
Bootstrapping scheduled tasks doesn't work when using the filesystem


dmcassel:
$ ./ml local bootstrap
    Bootstrapping your project into MarkLogic 9 on localhost...
    ERROR: ADMIN-NOSUCHDATABASE: (err:FOER0000) No such database filesystem
    See MarkLogic Server error log for more details.
    ERROR: ... Bootstrap FAILED
    ERROR: exit
    ERROR: ["deploy/lib/ml.rb:170:in `exit'", "deploy/lib/ml.rb:170:in `<main>'"]

This is because `setup:create-scheduled-task()` assumes a modules database. It's an easy fix; pull request soon.

dmcassel:
I saw this on MarkLogic 9, but the problem is independent of the version.

grtjn:
Fixed in dev..

PR #649, opened 2016-08-23
#648 check whether we're using the filesystem for modules

dmcassel:
#648

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -4120,12 +4120,17 @@ declare function setup:get-scheduled-task(
     admin:group-get-scheduled-tasks(
       $admin-config,
       $group-id)
+  let $modules-db :=
+    if ($task/gr:task-modules/@name eq "filesystem") then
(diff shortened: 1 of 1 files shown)


grtjn:
Bootstrap runs great, but wipe complains too. Can you look at that as well? :P

dmcassel:
Good catch, will fix.

dmcassel:
Fixed. Successfully wipes just the targeted scheduled tasks.

grtjn:
Using `./ml upgrade --fork=dmcassel --branch=issue-648` to test..

RobertSzkutak:
This only works if the task-modules database is set to filesystem. It still fails if the task-content database is set to filesystem.

grtjn:
Bootstrap and wipe failed before, but after upgrade both run fine..

dmcassel:
@RobertSzkutak we don't allow for content databases on the filesystem; it's just an issue for modules.

grtjn:
I'm merging!

Issue #656, opened 2016-09-02
support javascript suite-setup/setup/suite-teardown/teardown


bradmann:
add support for javascript setup and teardown scripts for unit tests.

dmcassel:
fixed in dev

dmcassel:
I updated the wiki to show the SJS setup and teardown filenames.

PR #657, opened 2016-09-02
#656 Support JavaScript setup/suite-setup/teardown/suite-teardown

bradmann:
JavaScript is awesome.

### DIFF
--- a/src/test/default.xqy
+++ b/src/test/default.xqy
@@ -124,9 +124,9 @@ declare function t:run-suite($suite as xs:string, $tests as xs:string*, $run-sui
         if ($ex/error:code = "XDMP-MODNOTFOUND" and
           fn:matches($ex/error:stack/error:frame[1]/error:uri/fn:string(), "/suite-setup.xqy$")) then
           try {
-            xdmp:invoke(fn:concat("suites/", $suite, "/suite-setup.sjs")),
+            xdmp:invoke(fn:concat("suites/", $suite, "/suiteSetup.sjs")),
             element t:test {
-              attribute name { "suite-setup.sjs" },
+              attribute name { "suiteSetup.sjs" },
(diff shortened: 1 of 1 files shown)


RobertSzkutak:
Pinging @grtjn and @dmcassel for an opinion as I'm not very familiar with the testing framework.

I feel like the best way to handle this would be with a variable in src/test/test-config.xqy to pick whether you want XQY or SJS. Could there ever be a scenario in which you would want to invoke both? Lastly, I think we would need sample SJS files in the repository to keep it newbie friendly.

On the surface, I think this absolutely benefits Roxy even if we would need to do some more work on it. I'll defer to someone else about whether our preferred process would be to first perfect the PR or merge and finish it later.

dmcassel:
I'll try to poke at it this week.

bradmann:
I've already coded a solution and submitted a pull request against dev.

dmcassel:
Rephrasing -- I'll check out your PR this week

dmcassel:
A couple things (nothing big):

- on [default.xqy line 456](https://github.com/bradmann/roxy/blob/77452b9cea62d8dd0b96fecec8659da01356f342/src/test/default.xqy#L456), there's a test to see whether a file should be considered a test. Since suite-setup.sjs and suite-teardown.sjs are not part of that test, they show up as unit tests and are executed as such (in addition to be run correctly as suite setup/teardown). 
- aren't filenames in JavaScript land more often camelCased? I'm thinking it would be more consistent to call these suiteSetup.sjs and suiteTeardown.sjs. (I know there isn't really a convention, I just think it makes sense to match the filename to what we do inside the file. Other opinions welcome, but absent consensus in the other direction, I'd like to see the filenames changed.)

Otherwise, this looks great.

bradmann:
changes made.

Issue #621, opened 2016-06-20
Passwords not bootstrapped correctly


grtjn:
appuser-password values containing `{` or `}` or `\` get mangled.

The first two are doubled, by changes I made myself, though I don't really see how it breaks exactly, see end of build_properties?

The latter just disappears, taking along the following character (if any?). Happens because gsub interprets replacement, and \ has special meaning. Tried to escape, but that didn't seem to help..

dmcassel:
fixed in dev

PR #658, opened 2016-09-08
Fix for issue #621

### LINKED ISSUE #622: Escape passwords on use, not upfront
Related to #621, when appuser-password contains characters like `' " < > & { }`, they get escaped for use in ml-config.xml, but that is done upfront, not at insert. This means that if you ask properties values internally or via `ml local info`, you get the escaped version..

RobertSzkutak:
#621 

I tested this with a variety of passwords including many special characters such as `{ } >< \ /`

Ran many tests across Roxy and couldn't find any evidence that this breaks anything.

Simple fix but took me a couple evenings to study the codebase to find it and a couple more to thoroughly test it.

Partially fixes #622 but characters like `>` and `<` are still shown as escaped. Probably will need a separate PR for that.

### DIFF
--- a/deploy/lib/util.rb
+++ b/deploy/lib/util.rb
@@ -145,7 +145,7 @@ class String
   end
 
   def xquery_safe
-    REXML::Text::normalize(self).gsub(/\{/, '{{').gsub(/\}/, '}}')
+    REXML::Text::normalize(self)
   end
 
   def xquery_unsafe


dmcassel:
Works for { and }, but doesn't look like it addresses <, >, \, or /.

RobertSzkutak:
Hi @dmcassel , in my testing I tried this as appuser-password "pm*5*#W>{[T,e}STojT5\" (no quotes) and it failed before but worked fine after the change. I tried a couple of other variants as well. Do you happen to have a password it didn't work for that I could try?

Issue #640, opened 2016-07-28
Support SJS controllers


grtjn:
ml extend only supports xqy extensions, and samples doesn't include sjs samples either..

prestonmcgowan:
/app/config/config.xqy
`
<request uri="^/api/print/here$" endpoint="/roxy/query-router.xqy">
	<http method="GET"/>
	<uri-param name="controller">test</uri-param>
	<uri-param name="language">sjs</uri-param>
	<uri-param name="func">print</uri-param>
	<uri-param name="format">json</uri-param>
</request>
<request uri="^/api/compute/area/(\d+)$" endpoint="/roxy/update-router.xqy">
	<http method="PUT"/>
	<uri-param name="controller">test</uri-param>
	<uri-param name="language">sjs</uri-param>
	<uri-param name="func">width</uri-param>
	<uri-param name="width">$1</uri-param>
	<uri-param name="format">json</uri-param>
</request>
`

/app/config/controllers/test.sjs
`var req = require("/roxy/lib/request.xqy");

module.exports = {

	print : function () {
		var rqst = xdmp.getRequestBody().toObject();
		xdmp.log(rqst.here);
		return "here";
	},
	area  : function () {
		var width = req.get("width");
		return width * width;
	}
};
`

I also have the test.sjs dumping the HTTP Request Body to show how you could get at the full contents of the request.

PR #673, opened 2016-10-19
#640: Support SJS with a language property

prestonmcgowan:
Add <uri-param name="language">sjs</uri-param> to your request
definition to make use of a SJS controller.
More comments added to Issue #640.

### DIFF
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -14,6 +14,13 @@
 xquery.dir=${basedir}/src
 
 #
+# What type of language is the default for the controllers?
+# sjs - Serverside JavaScript
+# xqy - XQuery (default)
+#
+controller-ext=xqy
+
+#
 # MarkLogic application servers exist inside a group. ML Instances start off
 # with a group called "Default".
 #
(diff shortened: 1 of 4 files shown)


dmcassel:
This requires adding the parameter on every call -- feels error prone. I like including this, but also controlling the default (xquery vs sjs). To do that, we would:

1. add a "controller-ext" property to default.properties, with a default value of "xqy"
2. add `declare variable $c:CTRL-EXT := ("@ml.controller-ext", $def:CTRL-EXT)[1];` to src/app/config/config.xqy
3. add `declare variable $CTRL-EXT := "xqy";` to src/roxy/config/defaults.xqy
4. in src/roxy/router.xqy: `declare variable $language as xs:string := req:get("language", $config:CTRL-EXT, "type=xs:string");

I'd have to test to make sure that's right, but I think that will do it. Then you could add 

    controller-ext=sjs

to your build.properties and default to SJS controllers, but still override if needed. Want me to send that change to your PR, or would you like to make that update?

prestonmcgowan:
Changes made and committed.
Updated sample code on #640.
Please review.

dmcassel:
Looks great. Thanks for the PR!

PR #674, opened 2016-10-19
Issue 563

dmcassel:
This PR makes unit tests work out of the box with the default user (ie, no longer requiring admin for tests to work). 

Question for reviewers: any concerns with the privileges that I've granted to the default user? 

- any-uri
- rest-reader
- xdmp:get-server-field
- xdmp:set-server-field

Three of those are required for Roxy's rewriter to work (see src/roxy/rewriter-lib.xqy rewrite-rules(); `eput:get-rest-options` requires rest-reader). 

The amps should be safe, in that the code they refer to isn't even deployed in production.

### DIFF
--- a/deploy/sample/ml-config.sample.xml
+++ b/deploy/sample/ml-config.sample.xml
@@ -458,6 +458,12 @@
       </collections>
       <privileges>
         <privilege>
+          <privilege-name>any-uri</privilege-name>
+        </privilege>
+        <privilege>
+          <privilege-name>rest-reader</privilege-name>
+        </privilege>
+        <privilege>
           <privilege-name>xdmp:value</privilege-name>
         </privilege>
         <privilege>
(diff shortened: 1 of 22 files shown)


RobertSzkutak:
Granting any-uri will definitely circumvent security at larger orgs and gov. Preferably could we just define the proper URI permissions?

deploy/sample/ml-config.sample.xml
```diff
@@ -458,6 +458,12 @@
       </collections>
       <privileges>
         <privilege>
+          <privilege-name>any-uri</privilege-name>
```
RobertSzkutak: Doesn't this one belong in the (optional) ${app-role}-unit-test ?

dmcassel: Ack. Yes, it does. I'll move that and run another comparison between the ml-config.xml I tested with and this file.

dmcassel:
@RobertSzkutak I moved the any-uri privilege.

RobertSzkutak:
That looks good! It worked fine last night when I tested it too. Merging..

PR #676, opened 2016-10-20
Fix #585

### LINKED ISSUE #585: Support for schema, namespaces, module locations and traces for groups

rlouapre:
#585 

New PR based in the latest dev branch

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -110,6 +110,11 @@ declare variable $group-settings :=
     <!--TODO: setting>audit-role-restriction</setting>
     <setting>audit-uri-restriction</setting>
     <setting>audit-user-restriction</setting-->
+    <setting min-version="7.0-0" function="add" keys="namespace-uri location">module-location</setting>
+    <setting min-version="7.0-0" function="add" keys="prefix namespace-uri">namespace</setting>
+    <setting min-version="7.0-0" function="add" keys="namespace-uri schema-location">schema</setting>
+    <setting min-version="7.0-0" function="add" keys="event-id">trace-event</setting>
+    <setting min-version="7.0-0" function="add" keys="namespace-uri">using-namespace</setting>
   </settings>;
 
 declare variable $host-settings :=
@@ -1105,17 +1110,16 @@ declare function setup:create-forests-from-config(
   let $hosts := admin:group-get-host-ids(admin:get-configuration(), $group-id)
   let $host-name as xs:string? := $forest-config/as:host-name[fn:string-length(fn:string(.)) > 0]
   let $host-id := if ($host-name) then xdmp:host($host-name) else ($hosts, $default-host)[1]
-  let $hostnr := fn:index-of($hosts, $host-id)
-  let $replica-names as xs:string* := $forest-config/as:replica-names/as:replica-name[fn:string-length(fn:string(.)) > 0]
-  let $replicas :=
-    $import-config/as:assignments/as:assignment[as:forest-name = $replica-names]
   return
     setup:create-forest(
       $forest-name,
       $data-directory,
       $host-id,
       if (fn:count($hosts) gt 1) then
-        setup:reassign-replicas($replicas, $hosts, $hostnr, $forest-name, 1, fn:false())
+        let $hostnr := fn:index-of($hosts, $host-id)
+        let $replica-names as xs:string* := $forest-config/as:replica-names/as:replica-name[fn:string-length(fn:string(.)) > 0]
+        let $replicas := $import-config/as:assignments/as:assignment[as:forest-name = $replica-names]
+        return setup:reassign-replicas($replicas, $hosts, $hostnr, $forest-name, 1, fn:false())
       else ()
     )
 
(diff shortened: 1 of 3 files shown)


rlouapre:
I had to comment out ```ruby require 'test/test_util' ``` see error below - then all tests passed:

```
C:\Users\rlouapre\Documents\GitHub\roxy>ml self-test
Loaded suite deploy/test/test_main
Started
F
================================================================================

Failure: test_xquery_safe_unsafe(TestProperties)
C:/Users/rlouapre/Documents/GitHub/roxy/deploy/test/test_util.rb:12:in `test_xqu
ery_safe_unsafe'
      9:   # test for issue #177
     10:   def test_xquery_safe_unsafe
     11:     assert_equal('secret', 'secret'.xquery_safe.xquery_unsafe)
  => 12:     assert_equal('crazypassword{123@#$%}},.<>', 'crazypassword{123@#$%}
},.<>'.xquery_safe.xquery_unsafe)
     13:   end
     14:
     15: end
<"crazypassword{123@\#$%}},.<>"> expected but was
<"crazypassword{123@\#$%},.<>">

diff:
? crazypassword{123@#$%}},.<>
================================================================================



Finished in 0.028002 seconds.
--------------------------------------------------------------------------------

1 tests, 2 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifica
tions
0% passed
--------------------------------------------------------------------------------

35.71 tests/s, 71.42 assertions/s
```

dmcassel:
@grtjn I think you're most familiar with the group functionality. Want to review this one?

rlouapre:
Any update on this PR?

dmcassel:
@rlouapre, @RobertSzkutak is planning to review this shortly

deploy/lib/xquery/setup.xqy
```diff
@@ -3136,6 +3140,59 @@ declare function setup:validate-group(
...
+  $setting as element()
+) {
+  let $group := $group-config/gr:group-name/fn:string(.)
+  let $setting-name := $setting/fn:string(.)
+  let $function-name := fn:concat("group-get-", $setting-name, "s")
+  let $keys := fn:tokenize($setting/@keys, " ")
+  let $existing :=
+    try
+    {
+      xdmp:eval(
```
RobertSzkutak: Look like the only reason you are evaling here is to return the function. Instead, couldn't you do this?:

```
let $config := admin:get-configuration()
let $groupid := admin:group-get-id($config, $group)
return xdmp:value(fn:concat('admin:', $function-name, '($config, $groupid)'))
```

Granted, I only played around with this a little in QConsole so let me know if I'm overlooking something here...

rlouapre: I suppose but the same technique: ```try{xdmp:eval}catch(){}``` is used in many places in the same module. That's the reason I choose to reuse it.

RobertSzkutak: Thanks for pointing that out. Took a quick look and saw the same. As you probably know, eval is slow and we should avoid using it unnecessarily wherever possible. CC @dmcassel @grtjn in case there's a reason to keep it that I'm not aware of.

Beyond this, your code was a pleasure to read through and I was quite impressed by the gymnastics you performed in lieu of missing admin:group-set* functions as well as complicated string concatenations you performed to make this code forward compatible with new versions of ML with little development effort. I'd have no problem merging this in with the eval removed.

dmcassel:
I believe there are three reasons that setup.xqy uses xdmp:eval: 

1. to run code against a different database
2. to run in a different transaction
3. for backward compatibility (if the code has function that was introduced in ML7, but is run against ML6, that would be a syntax error; wrapping the code in eval limits the scope of the problem, allowing the rest of the code to run. 

If none of those apply, then the eval should be removed (although if you have a good reason for using it, feel free to make the case). As of this writing, [Roxy supports MarkLogic versions 7 and up](https://github.com/marklogic/roxy/wiki/Supported-MarkLogic-versions).

RobertSzkutak:
Agreed @dmcassel .

However, he is correct that there are a handful of instances where xdmp:eval is used erroneously. One quick example is ODBC server construction: https://github.com/marklogic/roxy/blob/master/deploy/lib/xquery/setup.xqy#L3398

dmcassel:
That looks like category 3 (backwards compatibility). The `admin:odbc-server-create` function was added in ML6, so that xdmp:eval could be removed now.

RobertSzkutak:
@dmcassel I'm still not convinced eval should have been used in the first place there. Shouldn't the catch still function correctly regardless of whether eval is used or not? (Or does try-catch w/ eval behave differently pre-7?)

dmcassel:
The eval is needed because MarkLogic does a static check before running the code. If it sees a function it can't resolve, the code is rejected before execution starts. Paste this into QC:

    try {
      local:no-such-function()
    } catch($e) {
      "saved"
    }

The error doesn't get caught. Wrap it in an eval and it does.

dmcassel:
That said, anything that was added in MarkLogic 6 or earlier no longer needs to be eval-wrapped.

RobertSzkutak:
@dmcassel You could instead do:

`try {
  xdmp:value("local:no-such-function()")
} catch($e) {
  "saved"
}`

RobertSzkutak:
@dmcassel (Fully aware I'm going off on a tangent here...) I further think a better pattern for backwards compatibility would be :

    if(setup:at-least-version("6.0-2")) then
       do ODBC stuff
   else
      fn:error(xs:QName("UnsupportedVersion", "Your version of ML is not supported"))

That way you could avoid the try-catch too and further enhance performance.

dmcassel:
The `if (fn:not...` approach doesn't work, again because the code fails the static analysis and the if test is never run. Try this in QC:

    try {
      if (1 = 2) then
        local:no-such-function()
      else ()
    } catch ($e) {
      "saved"
    }

It still fails, even though the test will always be false. (This kind of test is sometimes used to trick the static analyzer into doing an update rather than a query statement.) 

The `xdmp:value` approach looks good. If that's faster, then for cases where we're only using `xdmp:eval` for backwards compatibility, we can switch. 

We should create a new issue to track this. The goal would be to revisit each eval and see whether 1) the code can be simply inlined; 2) if not, can `xdmp:value` be used instead of `xdmp:eval`. 

@rlouapre it looks like your case could use value rather than eval.

RobertSzkutak:
@dmcassel I brain-dumped and wrote that really poorly. In my if .. then example I should have specified that xdmp:value() would be called after then, thus allowing us to get around the static analysis and avoid try-catch too.

dmcassel:
@RobertSzkutak now you're talking. That approach sounds ideal (though we should still skip all that for functions added in ML6 or later).

RobertSzkutak:
Pushing into our next milestone due to merge conflicts and open performance enhancement recommendations

grtjn:
Sorry being rather late into the discussion. I think the concerns around xdmp:eval and such are valid, and very interesting, but are only side-ways relevant to this PR. Getting rid of xdmp:eval's is probably better off as a separate effort, so I would have went ahead and merged this PR, and opened a new ticket to get the xdmp:eval's removed (as far as possible)..

grtjn:
Unfortunately, this PR can no longer be merged. Not sure how much setup.xqy changed in the mean time..

grtjn:
@rlouapre, are you still around to look into resolving the merge conflict? Might be a simple git rebase, but not sure..

grtjn:
I'll merge manually..

grtjn:
Replaced by PR #842..

Issue #569, opened 2016-02-24
How to set "update-policy" property throught Roxy


romain-nio:
Hi,

I'm wondering if it's possible to set the "update-policy" property directly throught Roxy. I'm able to set it thourght API but I don't know how to do it with Roxy.

Thanks,
Romain.

grtjn:
If you mean `assignment-policy` on databases, then afraid not. I did see it when going through properties added in ML8, but this required a little more effort, not a straight-forward value assignment. Shouldn't be too difficult though, we have dealt with other properties that are similar in complexity..

grtjn:
Sorry, pushing away to allow release of 1.7.3..

rlouapre:
This is for optimistic locking. I think it is already possible [1]

Just add it in roxy/rest-api/config/properties.xml:

```xml
...
<update-policy>version-optional</update-version>
...
```

Assuming rest-* properties are correctly set ```deploy rest``` or ```deploy modules```

[1] - https://github.com/marklogic/roxy/wiki/REST-properties-format-change

grtjn:
Ah thnx, here the DMC docs on REST properties: http://docs.marklogic.com/REST/GET/v1/config/properties/[property-name]#config-properties

dmcassel:
I just sent a PR that adds this property to the sample properties, so that it will be present in the file and easier to edit.

PR #683, opened 2016-10-28
#569 adding update-policy to properties

dmcassel:


RobertSzkutak:
Looks good and works fine for me!

Issue #661, opened 2016-09-16
Allow execute_query with app_name set to ml.app-name


grtjn:
I changed execute_query_8 as shown below, and that allows me to at least invoke modules of current project with for instance:

```
  def replace_addresses
    country = find_arg(["--country"]) || "NL"
    lang = find_arg(["--language"]) || "Dutch"
    logger.info "Replacing customer addresses with #{country}/#{lang}.."
    r = execute_query(
      %Q{
        xquery version "1.0-ml";
        xdmp:invoke('/task/replace-addresses.xqy', (xs:QName("country"), "#{country}", xs:QName("lang"), "#{lang}"))
      },
      { :app_name => @properties["ml.app-name"] }
    )
    r.body = parse_body(r.body)
    logger.info r.body
  end
```

Patch for server_config.rb:

``` ruby
  def execute_query_8(query, properties = {})
    if properties[:app_name] != nil && properties[:app_name] != @properties["ml.app-name"]
      raise ExitException.new("Executing queries with an app_name (currently) not supported with ML8+")
    end

    headers = {
      "Content-Type" => "application/x-www-form-urlencoded"
    }

    params = {
      :xquery => query,
      :locale => LOCALE,
      :tzoffset => "-18000"
    }

    port = @qconsole_port
    if properties[:app_name] != nil
      port = @properties["ml.app-port"]
    end
    if properties[:db_name] != nil
      params[:database] = properties[:db_name]
    end

    r = go "#{@protocol}://#{@hostname}:#{port}/v1/eval", "post", headers, params

    raise ExitException.new(JSON.pretty_generate(JSON.parse(r.body))) if r.body.match(/\{"error"/)

    r
  end
```

grtjn:
I marked it as bug as the missing app_name feature is pretty severe. We could further improve on this, by discovering the rest-port from app_name. I think I did something similar with capturing of app-builder projects, which are rest-based too..

grtjn:
Even nicer would be having an invoke_query command, which takes modules-uri, and list of name/value pairs. Could make use of /v1/invoke, but would mean ML8+ support only. xcc invoke was never implemented from the looks of it..

grtjn:
The above mentioned patch only works for rest and hybrid projects, and app-level auth might interfere as well. A more robust method is this:

```ruby
  def execute_query_8(query, properties = {})
    if properties[:app_name] != nil && properties[:app_name] != @properties["ml.app-name"]
[...]

PR #684, opened 2016-11-01
Issue661

jamsilvia:
Ticket #661  : Expanded replica and forest generation so replicas are properly round-robin distributed and redistributed on scale-out, both for internal and defined forests.  Added methods to clean up the replicas that should be retired after their replacements are in sync replicating state.  This may take some time, so it is expected that the user would run clean-replicas at a later date.  To handle that, a state file for which replicas we are waiting on, and a state file for which replicas need to be deleted are stored.


RobertSzkutak:
Hi @jamsilvia could you please send this merge request to the dev branch for review? We don't accept merge requests directly into master.

Thank you!!

Issue #685, opened 2016-11-02
Shared Config for multiple applications sharing a single content db


jmeekhof:
The company I work for currently has several (roxy) applications that all shared a common content-db.

This makes sharing the content database configuration difficult among all the applications difficult.

I'd like to propose the following method to handle sharing portions of configuration among multiple applications:

Create a new configuration option (living in build.properties) that would instruct roxy to read another properties file, and apply those properties after default and build, but before the environment properties file.

This option would point to a properties file _outside_ the deploy directory. This directory would be a submodule within your project, thus allowing you to share whatever portions of the configuration you desire.

My initial investigation suggests that this check for the option would exist within server_config.rb in the 
`2679 def ServerConfig.properties(prop_file_location = @@path)`
`2680     default_properties_file = ServerConfig.expand_path("#{prop_file_location}/default.properties")`
`2681     properties_file = ServerConfig.expand_path("#{prop_file_location}/build.properties")`
`2682 `
`2683     raise ExitException.new("You must run ml init to configure your application.") unless File.exist?(properties_file)`
`2684 `
`2685     properties = ServerConfig.load_properties(default_properties_file, "ml.")`
`2686     properties.merge!(ServerConfig.load_properties(properties_file, "ml."))`
`2687     # Read properties, look for a config that would instruct us to read another optional properties file`
`2688     #`
`2689     # If that exists, then find file, and merge its properties`
`2690     `

RobertSzkutak:
In a nutshell you're asking for the directory where *.properties files are found to be configurable, right? Or is there more to it than that?

jmeekhof:
I wasn't asking where, so much as describing a problem and proposing a solution.

The function referenced is where I think I'd add the logic I'm proposing.

RobertSzkutak:
Spoke with Josh, his proposal is as follows:

1) Create four layers of properties files:

1. default.properties
2. build.properties
3. shared.properties
4. environment.properties

The reason for shared.properties is that he wants to store the properties within a git submodule which makes it impossible to keep them in the same directory as build.properties. Second, there may be different "groups" of projects in the future which makes this third layer desirable. Finally, you would have to specify in default.properties where build.properties should be pulled from which would defeat the purpose of never modifying default.properties.

Given this, I can't think of a way to avoid having the shared.properties.

2) Allow for customization of where there environment.properties and shared.properties are stored.

Seeking sign-off from @grtjn and @dmcassel as this would fundamentally shift how Roxy handles properties.

[...]

PR #686, opened 2016-11-03
Shared config

jmeekhof:
Pull request for issue #685

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2685,8 +2685,9 @@ private
     properties = ServerConfig.load_properties(default_properties_file, "ml.")
     properties.merge!(ServerConfig.load_properties(properties_file, "ml."))
 
+    #Look for optional shared_config, if it is set grab the properties from path relative to the root of the roxy project
     if properties['ml.shared_config'] then
-      shared_properties_file = ServerConfig.expand_path(properties['ml.shared_config'])
+      shared_properties_file = ServerConfig.expand_path("#{@@path}/../#{properties['ml.shared_config']}")
       properties.merge!(ServerConfig.load_properties(shared_properties_file))
     end
 
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2685,6 +2685,11 @@ private
     properties = ServerConfig.load_properties(default_properties_file, "ml.")
     properties.merge!(ServerConfig.load_properties(properties_file, "ml."))
 
+    if properties['ml.shared_config'] then
+      shared_properties_file = ServerConfig.expand_path(properties['ml.shared_config'])
+      properties.merge!(ServerConfig.load_properties(shared_properties_file))
+    end
+
     environments = properties['ml.environments'].split(",") if properties['ml.environments']
     environments = ["local", "dev", "prod"] unless environments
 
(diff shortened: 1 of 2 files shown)


deploy/lib/server_config.rb
```diff
@@ -2685,8 +2685,9 @@ def ServerConfig.properties(prop_file_location = @@path)
     properties = ServerConfig.load_properties(default_properties_file, "ml.")
     properties.merge!(ServerConfig.load_properties(properties_file, "ml."))
 
+    #Look for optional shared_config, if it is set grab the properties from path relative to the root of the roxy project
     if properties['ml.shared_config'] then
```
RobertSzkutak: I'm thinking we're also going to need to do these tests :

1) Does a shared properties file exist? (Could aid debugging)
2) Should a shared properties file be used for this environment? My understanding of your use case is that there's inevitably going to end up being two shared properties files applied to different environments respectively. So we'll need a mechanism to map environment properties to shared properties appropriately.

deploy/lib/server_config.rb
```diff
@@ -2685,8 +2685,9 @@ def ServerConfig.properties(prop_file_location = @@path)
...
-      shared_properties_file = ServerConfig.expand_path(properties['ml.shared_config'])
+      shared_properties_file = ServerConfig.expand_path("#{@@path}/../#{properties['ml.shared_config']}")
```
RobertSzkutak: I'm assuming we need to make the path configurable on this to fulfill your need to handle this via a submodule?

RobertSzkutak:
Great start! Will probably send you a PR this evening if you don't beat me to it..

deploy/lib/server_config.rb
jmeekhof: We could certainly add a file existence test, but that could cause issues with figuring out *if*  the shared config file is working. In other words, if the shared config file doesn't exist, and we're trying to use it, and the code silently jumps over the missing file, how do you know you've made a mistake? Sorry for the run on sentence.

RobertSzkutak: @jmeekhof I'm totally good with this after speaking with you about the use case.

RobertSzkutak:
Sent you a PR @jmeekhof with trivial documentation. If that looks good to you, I'll tap someone else to review before merging.

jmeekhof:
@RobertSzkutak I got your pull request, merged your change, and added a small comment of my own.

RobertSzkutak:
Looks good to me! @dmcassel or @grtjn could you also give this a quick review? Thanks!

dmcassel:
One small change and then I'm good for this to be merged:

> Config file loction is relative to the root of your roxy directory

Please change "roxy directory" to "deploy directory". For people using the MVC application structure, there is a src/roxy/ directory, so this could be confusing. (Might as well fix the "loction" typo at the same time.)

jmeekhof:
@dmcassel The way this is currently coded, its looking in the "root" of the application, not the deploy directory. When you say "deploy" directory, I think "deploy/". 

What is the best wording to use to avoid confusion, as I think "deploy" is also a bit confusing as well.

dmcassel:
In that case, I would refer to "the project root".

jmeekhof:
@dmcassel, @RobertSzkutak I've updated the wording to avoid inevitable confusion. I think it's ready to be merged!

RobertSzkutak:
Fine for me! Merging..

PR #689, opened 2016-11-15
Ticket661b

jamsilvia:
Ticket #661 : Expanded replica and forest generation so replicas are properly round-robin distributed and redistributed on scale-out, both for internal and defined forests. Added methods to clean up the replicas that should be retired after their replacements are in sync replicating state. This may take some time, so it is expected that the user would run clean-replicas at a later date. To handle that, a state file for which replicas we are waiting on, and a state file for which replicas need to be deleted are stored.

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -733,7 +733,7 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
 
     if r.body.match("error log")
       logger.error r.body
-      logger.error "... Bootstrap FAILED"
+      raise ExitException.new("... Bootstrap FAILED")
       return false
     else
       if r.body.match("(note: restart required)")
@@ -1055,6 +1055,29 @@ In order to proceed please type: #{expected_response}
   #
   def clear
     clean
+  end
+
+  #
+  # Install - Runs all steps needed to 'install' a Roxy application
+  #
+  def install
+    bootstrap
+    deploy_modules
+    deploy_schemas
+    if @properties["ml.triggers-db"] then
+      deploy_triggers
+    end
+    if @properties["ml.triggers-db"] and @properties["ml.data.dir"] and File.exist?(ServerConfig.expand_path("#{@@path}/pipeline-config.xml")) then
+      deploy_cpf
+    end
+    deploy_content
+  end
+
+  #
+  # Uninstall - an alternative command for wipe to complement install
+  #
+  def uninstall
+    wipe
   end
 
   #
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -652,17 +652,24 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
 
     internals = find_arg(['--replicate-internals'])
     if internals
+      dointernals = 'internals'
 
+      # Number of hosts
+      r = execute_query %Q{ fn:count(xdmp:hosts()) }
+      r.body = parse_body(r.body)
+
+      # check cluster size
       nr = find_arg(['--nr-replicas'])
       if nr
-        nr = nr.to_i
+        if nr == "max" or nr == "MAX"
+          nr = r.body.to_i - 1
+        else
+          nr = nr.to_i
+        end
       else
         nr = 2
       end
 
-      # check cluster size
-      r = execute_query %Q{ fn:count(xdmp:hosts()) }
-      r.body = parse_body(r.body)
       raise ExitException.new("Increase nr-replicas, minimum is 1") if nr < 1
       raise ExitException.new("Adding #{nr} replicas to internals requires at least a #{nr + 1} node cluster") if r.body.to_i <= nr
 
@@ -672,16 +679,12 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
       assigns = ''
       internals = @properties['ml.system-dbs'].split ','
       internals.each do |db|
-        repnames = ''
-        repassigns = ''
-        (1..nr).each do |i|
-          repnames = repnames + %Q{
-                <replica-name>#{db}-rep#{i}</replica-name>}
-          repassigns = repassigns + %Q{
+        repnames = %Q{
+            <replica-name>#{db}-rep1</replica-name>}
+        repassigns = %Q{
             <assignment>
-              <forest-name>#{db}-rep#{i}</forest-name>
+              <forest-name nr-replicas="#{nr}">#{db}-rep1</forest-name>
             </assignment>}
-        end
 
         assigns = assigns + %Q{
 
@@ -714,18 +717,19 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
       }
       logger.debug config
     else
+      dointernals = ''
       logger.info "Bootstrapping your project into MarkLogic #{@properties['ml.server-version']} on #{@hostname}..."
       config = get_config
     end
 
     apply_changes = find_arg(['--apply-changes'])
 
-    if apply_changes == nil
-      apply_changes = ""
+    if apply_changes == nil or apply_changes == ""
+      apply_changes = "all"
     end
 
     setup = File.read(ServerConfig.expand_path("#{@@path}/lib/xquery/setup.xqy"))
-    r = execute_query %Q{#{setup} setup:do-setup(#{config}, "#{apply_changes}")}
+    r = execute_query %Q{#{setup} setup:do-setup(#{config}, "#{apply_changes},#{dointernals}")}
     logger.debug "code: #{r.code.to_i}"
 
     r.body = parse_body(r.body)
@@ -744,6 +748,73 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
       logger.info "... Bootstrap Complete"
       return true
     end
+  end
+
+  def clean_replicas_state
+    internals = find_arg(['--internal-replicas'])
+
+    if internals == nil
+      internals = ''
+      logger.info "Cleaning application forest decommissioned replica state"
+      config = get_config
+    else
+      logger.info "Cleaning interal forest decommissioned replica state"
+      internals = 'internals'
+      config = get_config
+    end
+
+    setup = File.read(ServerConfig.expand_path("#{@@path}/lib/xquery/setup.xqy"))
+    r = execute_query %Q{#{setup} setup:do-clean-replicas-state(#{config}, "#{internals}")}
+
+    if r.body.match("error log")
+      logger.error r.body
+      logger.error "... Cleaning replicas FAILED"
+      return false
+    end
+
+    logger.info r.body
+    logger.info "... Cleaning replicas Complete"
+    return true
+  end
+
+  def clean_replicas
+    internals = find_arg(['--internal-replicas'])
+
+    if internals == nil
+      internals = ''
+      logger.info "Cleaning application forest decommissioned replicas, if ready."
+      config = get_config
+    else
+      logger.info "Cleaning interal forest decommissioned replicas, if ready."
+      internals = 'internals'
+      config = get_config
+    end
+
+    setup = File.read(ServerConfig.expand_path("#{@@path}/lib/xquery/setup.xqy"))
+    r = execute_query %Q{#{setup} setup:do-clean-replicas(#{config}, "#{internals}")}
+    logger.debug "code: #{r.code.to_i}"
+
+    r.body = parse_body(r.body)
+    logger.debug r.body
+
+    if r.body.match("error log")
+      logger.error r.body
+      logger.error "... Cleaning replicas FAILED"
+      return false
+    end
+    if r.body.match("Replicas not ready")
+      logger.error r.body
+      return false
+    end
+    if r.body.match("nothing to do")
+      logger.error r.body
+      logger.info "No replicas were found to be retired.  Nothing to do."
+      return false
+    end
+
+    logger.info r.body
+    logger.info "... Cleaning replicas Complete"
+    return true
   end
 
   def wipe
@@ -882,8 +953,8 @@ In order to proceed please type: #{expected_response}
 
       wipe_changes = find_arg(['--apply-changes'])
 
-      if wipe_changes == nil
-        wipe_changes = ""
+      if wipe_changes == nil or wipe_changes == ""
+        wipe_changes = "all"
       end
 
       r = execute_query %Q{#{setup} setup:do-wipe(#{config}, "#{wipe_changes}")}
@@ -1044,6 +1115,10 @@ In order to proceed please type: #{expected_response}
         clean_cpf
       when 'triggers'
         clean_triggers
+      when 'replicas'
+        clean_replicas
+      when 'replicas-state'
+        clean_replicas_state
       else
         raise HelpException.new("clean", "Invalid WHAT")
     end
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -661,13 +661,13 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
       # check cluster size
       nr = find_arg(['--nr-replicas'])
       if nr
-        if nr == "max" or nr == "MAX"
+        if nr.downcase == "max"
           nr = r.body.to_i - 1
         else
           nr = nr.to_i
         end
       else
-        nr = 2
+        nr = 1
       end
 
       raise ExitException.new("Increase nr-replicas, minimum is 1") if nr < 1
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1640,7 +1640,7 @@ private
           File.open("#{path}", 'wb') { |file| file.write(r.body) }
         end
       end
-    else
+    elsif @properties['ml.server-version'] == '7'
       # In ML7, the response is JSON
       # [
       #  {"qid":null, "type":"string", "result":"\/"},
@@ -1657,6 +1657,42 @@ private
         FileUtils.mkdir_p(parentdir) unless File.exists?(parentdir)
         if ! uri.end_with?("/")
           File.open("#{path}", 'wb') { |file| file.write(r.body) }
+        end
+      end
+    else
+      # ML8, we're using /v1/eval, so we get a multi-part response
+      uris = parse_body(dirs.body)
+      uris.split(/\r?\n/).each do |uri|
+        if ! uri.end_with?("/")
+
+          r = execute_query %Q{
+            fn:doc("#{uri}")
+          },
+          { :db_name => target_db }
+
+          delimiter = r.body.split("\r\n")[1].strip
+          parts = r.body.split(delimiter)
+
+          # The first part will always be an empty string. Just remove it.
+          parts.shift
+          # The last part will be the "--". Just remove it.
+          parts.pop
+
+          # Get rid of part headers
+          parts = parts.map{ |part|
+            sections = part.split("\r\n\r\n");
+            sections.slice(1, sections.length).join("\r\n\r\n")
+          }
+
+          # Return all parts as one long string, like we were used to.
+          parts = parts.join().chomp("\r\n")
+
+          path = "#{target_dir}#{uri}"
+          parentdir = File.dirname path
+          FileUtils.mkdir_p(parentdir) unless File.exists?(parentdir)
+          if ! uri.end_with?("/")
+            File.open("#{path}", 'wb') { |file| file.write(parts) }
+          end
         end
       end
     end
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2252,6 +2252,8 @@ private
       sub_me.each do |k,v|
         if v.match(/\$\{basedir\}/)
           sub_me[k] = ServerConfig.expand_path(v.gsub("${basedir}", Dir.pwd))
+          matches = v.scan(/\$\{([^}]+)\}/)
+          needs_rescan = true if matches.length > 1
         else
           matches = v.scan(/\$\{([^}]+)\}/)
           if matches.length > 0
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -1302,7 +1302,6 @@ declare function setup:save-cleanup-state( $import-config as element(configurati
                  'sec:get-role-names( xdmp:get-current-roles() )', 
                  (), 
                  <options xmlns="xdmp:eval"><database>{xdmp:security-database()}</database></options> )
-let $_ := xdmp:log( "EVAL ROLE: " || $user-roles )
   let $perms :=
       if( "admin" = $user-roles ) then
         (
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -1297,12 +1297,26 @@ declare function setup:save-cleanup-state( $import-config as element(configurati
           $replicating-map-file-internal
         else
           $replicating-map-file
+  let $user-roles := 
+      xdmp:eval( 'import module namespace sec="http://marklogic.com/xdmp/security" at "/MarkLogic/security.xqy";' ||
+                 'sec:get-role-names( xdmp:get-current-roles() )', 
+                 (), 
+                 <options xmlns="xdmp:eval"><database>{xdmp:security-database()}</database></options> )
+let $_ := xdmp:log( "EVAL ROLE: " || $user-roles )
+  let $perms :=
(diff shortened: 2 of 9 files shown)


deploy/lib/server_config.rb
```diff
@@ -652,17 +652,24 @@ def bootstrap
...
 
+      # Number of hosts
+      r = execute_query %Q{ fn:count(xdmp:hosts()) }
+      r.body = parse_body(r.body)
+
+      # check cluster size
       nr = find_arg(['--nr-replicas'])
       if nr
-        nr = nr.to_i
+        if nr == "max" or nr == "MAX"
```
RobertSzkutak: Could you instead do something along the lines of `if nr.downcase == "max"` ?

deploy/lib/server_config.rb
```diff
@@ -652,17 +652,24 @@ def bootstrap
...
+          nr = r.body.to_i - 1
+        else
+          nr = nr.to_i
+        end
       else
         nr = 2
```
RobertSzkutak: Wouldn't 1 be a better default number of replicas since that is the minimum number of replicas allowed? (according to line 673..)

deploy/lib/server_config.rb
```diff
@@ -672,16 +679,12 @@ def bootstrap
       assigns = ''
       internals = @properties['ml.system-dbs'].split ','
       internals.each do |db|
-        repnames = ''
-        repassigns = ''
-        (1..nr).each do |i|
-          repnames = repnames + %Q{
-                <replica-name>#{db}-rep#{i}</replica-name>}
-          repassigns = repassigns + %Q{
```
RobertSzkutak: If I'm reading this correctly, and I could be wrong here, are you only allowing one replica now?

jamsilvia: The old way the code worked was to generate the replicas in the server_config.rb.  I moved the code to generate the replicas into the setup.xqy.  The config now just sends the count for the number of replicas across.

deploy/lib/server_config.rb
jamsilvia: Concur.  Updated.

deploy/lib/server_config.rb
jamsilvia: Concur.  Updated.  (That was left over from testing - thank you!)

deploy/lib/xquery/setup.xqy
```diff
@@ -1067,24 +1244,93 @@ declare function setup:validate-mimetypes($import-config as element(configuratio
...
+  let $which-replicating-map-file := 
+        if( $do-internals ) then
+          $replicating-map-file-internal
+        else
+          $replicating-map-file
+
+  (: Write the delete maps and the replicating maps for use when delete old replicas is done :)
+  return
+  (
+    xdmp:document-insert( $which-delete-map-file, document { $delete-map } ),
```
RobertSzkutak: Could you explicitly force admin permissions on these to prevent users from accidentally deleting them?

jamsilvia: I checked in an update to setup.xqy that hopefully addresses your concern.  I didn't create and permission if the default wasn't there - at that point, I wouldn't know what permission to explicitly use.  So I stopped at default-permission.

RobertSzkutak:
Great! This looks okay to me. @grtjn could you give it final approval?

jamsilvia:
Any word on this @grtjn ?

jamsilvia:
@RobertSzkutak any word on this moving to closed?  Just wondering if you are still waiting on something from me, or if it is in the works.  Thank you!

RobertSzkutak:
I am okay with merging this. However, as a best practice, we usually have at least two people from our team review larger commits. @dmcassel , would you be able to take a look at this?

deploy/lib/xquery/setup.xqy
```diff
@@ -1067,24 +1244,106 @@ declare function setup:validate-mimetypes($import-config as element(configuratio
...
+    xdmp:document-insert( $which-replicating-map-file, document { $replicating-map }, $perms )
+  )
+};
+
+declare function setup:create-forests($import-config as element(configuration), $do-internals as xs:boolean) as item()*
+{
+  let $_ := setup:initialize-cleanup-state( $import-config, $do-internals )
+
+  let $return :=
+      for $db-config in setup:get-databases-from-config($import-config)
```
dmcassel: please indent 2

deploy/lib/xquery/setup.xqy
```diff
@@ -1093,32 +1352,38 @@ declare function setup:validate-forests($import-config as element(configuration)
...
-      $data-directory,
-      $host-id,
-      if (fn:count($hosts) gt 1) then
-        setup:reassign-replicas($replicas, $hosts, $hostnr, $forest-name, 1, fn:false())
-      else ()
-    )
 
+  (: Get the assignment entries for each primary forest ID associated with this database :)
+  for $forest-config in setup:get-database-forest-configs($import-config, $database-name)
+    let $forest-name as xs:string := $forest-config/as:forest-name[fn:string-length(fn:string(.)) > 0]
```
dmcassel: restore project's indentation

deploy/lib/xquery/setup.xqy
```diff
@@ -1131,75 +1396,160 @@ declare function setup:validate-forests-from-config(
   let $data-directory as xs:string? := $forest-config/as:data-directory[fn:string-length(fn:string(.)) > 0]
   let $host-name as xs:string? := $forest-config/as:host-name[fn:string-length(fn:string(.)) > 0]
   let $replica-names as xs:string* := $forest-config/as:replica-names/as:replica-name[fn:string-length(fn:string(.)) > 0]
-  return
+  let $a :=
```
dmcassel: `$a` doesn't do anything to improve the code's self-documentation. Restore to returning the result of `setup:validate-forest` directly.

deploy/lib/xquery/setup.xqy
```diff
@@ -1131,75 +1396,160 @@ declare function setup:validate-forests-from-config(
...
-    setup:create-forest(
-      $new-forest-name,
-      $data-directory,
-      $host,
-      if (fn:count($hosts) gt 1) then
-        setup:reassign-replicas($replicas, $hosts, $hostnr, $forest-name, $forestnr, fn:true())
+
+  (: Get the assignment entries for each primary forest ID associated with this database :)
+  for $forest-config in setup:get-database-forest-configs($import-config, $database-name)
+    let $forest-name as xs:string := $forest-config/as:forest-name[fn:string-length(fn:string(.)) > 0]
```
dmcassel: restore project standard indenting

dmcassel:
I don't have much experience working with replicas, so I'm mostly focusing on coding style and standards. Quick question though: if someone has an existing cluster with replica forests, will bootstrap and wipe work correctly on that cluster? In other words, would running this code result in orphaned forests? If the code relies existing replica forests having particular names, that seems likely.

dmcassel:
In earlier comments, you pointed out that you'd moved the replica forest generation from server_config.rb to setup.xqy. There is a downside to this -- with that code in server_config.rb, users can override with app_specific.rb if they want a different arrangement. I guess that would no longer be possible. @RobertSzkutak, would that be bad?

deploy/lib/xquery/setup.xqy
```diff
@@ -1131,75 +1396,160 @@ declare function setup:validate-forests-from-config(
...
+  replacements have gone to a sync replicating state to ensure there is not a time that at least one replica
+  is in a ready state.
+:)
+declare function setup:mark-old-replicas-for-delete(
+  $forest-name as xs:string
+) as xs:string*
+{
+  let $admin-config := admin:get-configuration()
+  let $existing-replicas := 
+      if( admin:forest-exists( $admin-config, $forest-name ) ) then
```
dmcassel: please indent 2

deploy/lib/xquery/setup.xqy
```diff
@@ -1131,75 +1396,160 @@ declare function setup:validate-forests-from-config(
...
+        let $forest-id := admin:forest-get-id($admin-config, $forest-name)
+        return admin:forest-get-replicas($admin-config, $forest-id)
       else ()
-    )
+
+  (: Loop over existing forest IDs and keep track of the forest names.  Any not reused will be marked for retirement :)
+  let $_ :=
+      for $rep-id in $existing-replicas
```
dmcassel: please use project-standard indenting

deploy/lib/xquery/setup.xqy
```diff
@@ -1131,75 +1396,160 @@ declare function setup:validate-forests-from-config(
...
 
+declare function setup:get-assigned-replicas(
+  $replicas as element(as:assignment)*,
+  $hosts as xs:unsignedLong+,
+  $hostnr as xs:int,
+  $forest-name as xs:string,
+  $forestnr as xs:int?
+) as xs:string*
+{
+  let $ret := setup:reassign-replicas( $replicas, $hosts, $hostnr, $forest-name, $forestnr, fn:false() )/as:forest-name/fn:string()
```
dmcassel: avoided the unnecessary let, just call the function

deploy/lib/xquery/setup.xqy
```diff
@@ -1131,75 +1396,160 @@ declare function setup:validate-forests-from-config(
...
-              "-"
-            )
-          else ()
-        ), "-")
-      }</as:forest-name>,
-      <as:host-name>{$replica-host-name}</as:host-name>,
-      $replica/node() except ($replica/as:forest-name, $replica/as:host-name)
-  }
+
+    (: If a limit on the number of replicas was specified, then use it. :)
```
dmcassel: restore project-standard indentation

deploy/lib/xquery/setup.xqy
```diff
@@ -1209,25 +1559,25 @@ declare function setup:validate-forests-from-count(
...
-    let $replica-name as xs:string := ($replica/as:forest-name[fn:string-length(fn:string(.)) > 0], fn:concat($forest-name, '-replica'))[1]
-    return
-      fn:string-join(($replica-name, fn:format-number(xs:integer($hostnr), "000"), xs:string($forestnr)), "-")
-  return
-    setup:validate-forest(
-      $forest-name,
-      $data-directory,
-      $host,
-      $replica-names)
+    let $forest-name as xs:string := $forest-config/as:forest-name[fn:string-length(fn:string(.)) > 0]
```
dmcassel: indentation

deploy/lib/xquery/setup.xqy
```diff
@@ -1434,9 +1784,12 @@ declare function setup:attach-forests-by-config(
   $db-config as element(db:database),
   $database-name as xs:string) as item()*
 {
-  for $forest-ref in $db-config/db:forests/db:forest-id
-  return
-    setup:attach-database-forest($database-name, fn:data($forest-ref/(@name|text())))
+  let $group-id := setup:get-group($db-config)
+  let $hosts := admin:group-get-host-ids(admin:get-configuration(), $group-id)
+  for $forest-config in setup:get-database-forest-configs($import-config, $database-name)
+    let $forest-name as xs:string := $forest-config/as:forest-name[fn:string-length(fn:string(.)) > 0]
```
dmcassel: indentation

deploy/lib/xquery/setup.xqy
```diff
@@ -1453,22 +1806,24 @@ declare function setup:attach-forests-by-count($db-config as element(db:database
...
   let $database-name := setup:get-database-name-from-database-config($db-config)
-  for $host at $hostnr in admin:group-get-host-ids(admin:get-configuration(), $group-id)
-  let $hostname := xdmp:host-name($host)
-  for $forestnr in (1 to setup:get-forests-per-host-from-database-config($db-config))
-  let $forest-name := fn:string-join(($database-name, fn:format-number(xs:integer($hostnr), "000"), xs:string($forestnr)), "-")
-  return
-    setup:attach-database-forest($database-name, $forest-name)
+  let $hosts := admin:group-get-host-ids(admin:get-configuration(), $group-id)
+  for $host at $hostnr in $hosts
+    let $hostname := xdmp:host-name($host)
```
dmcassel: indentation

deploy/lib/xquery/setup.xqy
```diff
@@ -1453,22 +1806,24 @@ declare function setup:attach-forests-by-count($db-config as element(db:database
...
+      return
+        setup:attach-database-forest($database-name, $forest-name)
 };
 
 declare function setup:validate-attached-forests-by-count($db-config as element(db:database))
 {
   let $group-id := setup:get-group($db-config)
   let $database-name := setup:get-database-name-from-database-config($db-config)
-  for $host at $hostnr in admin:group-get-host-ids(admin:get-configuration(), $group-id)
+  let $hosts := admin:group-get-host-ids(admin:get-configuration(), $group-id)
```
dmcassel: why this change?

RobertSzkutak:
Great catches @dmcassel , thanks!! To answer your question, I actually think it's highly unlikely that moving picking replica names out of server_config and into setup.xqy will matter since it was done in def bootstrap. If you've touched that code, then you'd have to have overridden def bootstrap by pretty closely copy and pasting the entire function. 

His forest name creation is done using the hostname in the forest name which is pretty standard in most large enterprise deployments I've seen. Only way to get those hostnames requires using xqy functions. While you could throw in an extra eval and pick the names in Ruby, I'm really not bothered by doing it in xqy.

Regarding orphaned forests, doesn't that already happen with Roxy currently?

dmcassel:
RE orphaned forests, yes, every time we change how they are named. For wipe, we could look up what forests are attached to a database and remove those (don't remember if that's what's currently happening). 

We're planning to do a release by 2017-01-31; do we want to include this, or let it marinate on the dev branch for a bit to make sure it works well? (Rob, you'll be more impacted by this change than me, so I'm willing to go with what you think.)

RobertSzkutak:
@dmcassel At least in regards to what you pointed out in your review, it should sit until those minor issues are resolved. If resolved before 1/31 I'm okay with merging this. I believe it adds quite a bit of value. In the past, I've actually avoided using Roxy for cluster-wide forest configuration and instead done it by hand. In fact, part of my motivation in creating `--apply-changes` was being able to bootstrap without touching the forests. In my testing, this has been a huge improvement and makes me feel a safer about bootstrapping forests on a cluster.

deploy/lib/xquery/setup.xqy
```diff
@@ -655,39 +723,44 @@ declare function setup:do-wipe($import-config as element(configuration)+, $optio
...
-              let $hosts := admin:group-get-host-ids(admin:get-configuration(), $group-id)
-              for $host at $hostnr in $hosts
-              for $forestnr in (1 to $forests-per-host)
-              for $replica in $import-config/as:assignments/as:assignment[as:forest-name = $replica-names]
-              let $replica-name as xs:string := ($replica/as:forest-name[fn:string-length(fn:string(.)) > 0], fn:concat($forest-name, '-replica'))[1]
-              return
-                fn:string-join(($replica-name, fn:format-number(xs:integer($hostnr), "000"), xs:string($forestnr)), "-")
-            else ()
-          )
+            let $forest-name := $assignment/as:forest-name
```
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
```diff
@@ -945,6 +1018,114 @@ declare function setup:do-wipe($import-config as element(configuration)+, $optio
...
+  let $_ := setup:initialize-cleanup-state( $import-config, $do-internals )
+
+  return
+    try
+    {
+      if( map:count( $delete-map ) ) then
+        (: Loop over the replicas we are waiting for and check state :)
+        let $reps-waiting :=
+            for $rep in map:keys( $replicating-map )
+              let $rep-id := xdmp:forest( $rep )
```
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
```diff
@@ -945,6 +1018,114 @@ declare function setup:do-wipe($import-config as element(configuration)+, $optio
...
+
+        (: If all replicas are sync'd, then we can remove the decommissioned replicas :)
+        return
+          if( fn:count( $reps-waiting ) = 0 ) then
+            let $_ := xdmp:log( "All new replicas are sync'd - cleaning decommissioned replicas" )
+            let $admin-config := admin:get-configuration()
+            let $updated-config := map:map() ! ( map:put( ., "admin-config", $admin-config ), . )
+            let $cleaned :=
+                for $del-rep in map:keys( $delete-map ) 
+                  (: First, get ID of the replica to delete :)
```
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
```diff
@@ -945,6 +1018,114 @@ declare function setup:do-wipe($import-config as element(configuration)+, $optio
...
+(:
+  Cleanup scale-out replica state files.
+:)
+declare function setup:do-clean-replicas-state($import-config as element(configuration)+, $options as xs:string) as item()*
+{
+  let $optionsMap := setup:parse-options( $options )
+  let $do-internals := map:contains( $optionsMap, "internals" )
+
+  let $which-delete-map-file := 
+        if( $do-internals ) then
```
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
```diff
@@ -1067,24 +1244,106 @@ declare function setup:validate-mimetypes($import-config as element(configuratio
...
+:)
+declare function setup:initialize-cleanup-state( $import-config as element(configuration), $do-internals as xs:boolean )
 {
-  for $db-config in setup:get-databases-from-config($import-config)
-  let $database-name := setup:get-database-name-from-database-config($db-config)
-  let $forests-per-host as xs:integer? := $db-config/db:forests-per-host
-  where fn:not($database-name = 'filesystem')
+  (: Get which files to use, based on whether using internals :)
+  let $which-delete-map-file := 
+        if( $do-internals ) then
```
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
```diff
@@ -1067,24 +1244,106 @@ declare function setup:validate-mimetypes($import-config as element(configuratio
...
+};
+
+(:
+  Save the scale-out cleanup state files to the database.  These are later used on subsequent runs of roxy
+  when a "clean replicas" is issued.
+:)
+declare function setup:save-cleanup-state( $import-config as element(configuration), $do-internals as xs:boolean )
+{
+  let $which-delete-map-file := 
+        if( $do-internals ) then
```
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: I'll fix it up.  No problem.

deploy/lib/xquery/setup.xqy
jamsilvia: Agree!  No idea why that is still there.  Will remove.

deploy/lib/xquery/setup.xqy
jamsilvia: Similar to above - maybe leftover from debug logging.  Will remove.

deploy/lib/xquery/setup.xqy
jamsilvia: No longer needed - will change back to the original.

jamsilvia:
For the fixes - I will get them in by early next week, no problem.

Re the orphaned forests - is it the forests themselves you are speaking of, or the replica forests?  I have added the "clean replicas" in order to handle any replicas being renamed/reassigned after the scale-out.

RobertSzkutak:
@jamsilvia Basically we were wondering what happens if you have forests and/or replica forests within a cluster that don't follow your naming convention. Could those become orphaned (not belonging to any database) on bootstrap or wipe? There are some issues with orphaned forests in Roxy prior to your commit.  (I do plan to test this when I get a spare moment, of course)

jamsilvia:
@RobertSzkutak Ok, let me break this down and see if I can cover the bases.  (ok, this is getting long):

Lets start with running wipe:
- When it comes to forests and replicas, I think this will act just like it did with older versions of Roxy.  I tried to stay with the existing way that Roxy dealt with this - which was to generate the forest names on the fly (vs looking up the set of existing forests for the database).  I'm not sure why this manner of cleanup was chosen, so I stayed with it in case there was some reasoning I was not aware of.  (But I think we can change this going forward somewhat easily).  This same concept applies to both the forests and replicas on wipe.

Running bootstrap:
- Running bootstrap should also be very similar to how it ran before.  Again, I tried to limit my changes to how things were done before while making it clearer on the forest and replica layout.
- Bootstrap shouldn't cause and orphaned forests itself, but does create additional forests with the new naming (which leads to the orphaned forests when trying to wipe).
- Replicas, however, are different now than they were before.  Since scale-out really affects the replica layout, replicas that are existing when re-running a bootstrap should not end up orphaned - if this concept:

The concept is this (assume all running under the new Roxy): You've run bootstrap once, and now you have a need to scale out your cluster.  You add more node(s) and want to include them as D-nodes.  So you re-run Roxy bootstrap, with the number of hosts bumped up.  If you really do it right, then you run it once again, with "--replicate-internals" to redistribute the replicas for the internal databases too.  At this point, new forests and replicas exist, and no "old" replicas are deleted.  Roxy has added replicas under the new naming and these replicas are distributed across the nodes appropriate for the new node count.  All of the old replicas are marked internally to be removed (by old replicas, I mean any replica that is reassigned to a different host due to scale out, or any replica manually created that doesn't match the naming).  At this point, MarkLogic is "async replicating" to the new replicas and the old ones are still sitting around.  You can try to run "clean replicas", but at this time Roxy should tell you it can't run until all replicas are "sync replicating".  Once the new replicas are "sync replicating" we know the new replicas are ready to go and are valid.  If "clean replicas" is run at this point (once for application replicas and once for internal replicas), Roxy will remove the old replicas and the resulting system looks like a bootstrap was run against a clean system.

This brings up (at least) two concerns:

1. I now possibly have many more replicas than originally planned - disk space may be a concern if I was already close to the limit.  Hopefully the system isn't managed poorly to be in this state, as having both replicas exist at the same time is safer; it always provides a valid replica in case of a failure during the async replicating time.  The future plan for this was to add in a flag that you could override this to force a clean of the old replicas immediately (at the expense of a window of time that the database does not have active full replicas).  Only recommended if you have done a DB backup immediately before.

2. What if I initially built out this cluster with an older version of Roxy?  In this case, there are some side effects.  First is that new forests will be created on scale-out with the new name (likewise with the replicas).  They won't be orphaned, but they would require a manual step to retire them, wait until they are cleared and data moved to the new forest, and then a manual delete.  Nothing is orphaned (unless wipe is performed), but there are now extra forests around that are left untouched.  I did not code the changes to automatically handle this case.  Considering it again, I suppose I could add in code to find the old names and simply rename the forest to the new naming, skipping the create.

What did I miss?  Let me know what I didn't cover - I want to make sure we have the use cases covered and a plan with how best to handle some of the complications typical users will come across.  If we need changes in before merge, let me know.

Thank you!

jamsilvia:
@dmcassel Updated based on your review.

Issue #568, opened 2016-02-19
Answering ml new warning with enter breaks..


grtjn:
Running ml new from within Roxy folder is not recommended, the ml script detects that, and shows a warning. If you hit enter to accept the default you get something like:

```
$ ./ml new xxx
Running ml new from within a Roxy project is not recommended. Continue? [y/N] 

./ml: line 49: [: !=: unary operator expected

Creating new Application: xxx...
Cloning into 'xxx'...
remote: Counting objects: 4442, done.
remote: Total 4442 (delta 0), reused 0 (delta 0), pack-reused 4442
Receiving objects: 100% (4442/4442), 7.83 MiB | 1.26 MiB/s, done.
Resolving deltas: 100% (2596/2596), done.
Checking connectivity... done.
Required option --server-version=[version] not specified with valid value.

  What is the version number of the target MarkLogic server? [5, 6, 7, or 8]
ERROR: No such file or directory - xxx
ERROR: [".../xxx/deploy/lib/server_config.rb:286:in `gets'", ".../xxx/deploy/lib/server_config.rb:286:in `gets'", ".../xxx/deploy/lib/server_config.rb:286:in `prompt_server_version'", ".../xxx/deploy/lib/server_config.rb:188:in `init'", "deploy/lib/ml.rb:126:in `<main>'"]
```

$response needs to be wrapped in double quotes in the ml-script..

grtjn:
I vaguely recall this was a one-line change. Pushing away nonetheless..

PR #700, opened 2016-11-18
Issue 568

RobertSzkutak:
#568


RobertSzkutak:
(Not really sure why there was a whitespace change. I am using the recommended text editor settings..)

dmcassel:
Does this need to be addressed in ml.bat as well?

RobertSzkutak:
Nope, ml.bat works correctly

Issue #694, opened 2016-11-16
property values using ${basedir} not rescanned for additional substitutions


kford-oconnors:
When both ${basedir} and a second property value are used in a third property, the second property 
name is not replaced by its value in the value of the third property.

Because ${basedir} is a special case in substitute_properties ([server_config.rb#L2246](https://github.com/marklogic/roxy/blob/1-7-4/deploy/lib/server_config.rb#L2246)) the lines where it occurs are not rescanned for additional properties.

Example: 
`
config.file=${basedir}/deploy/ml-config-${environment}.xml
`
config.file > [path to app directory]/deploy/ml-config-${environment}.xml

Roxy version: 1.7.4

RobertSzkutak:
Thanks for the PR to fix this, @kford-oconnors !

PR #702, opened 2016-11-18
Issue 694: rescan ${basedir} properties

kford-oconnors:
#694 

rescan properties containing ${basedir} when another property is present


RobertSzkutak:
Worked like a charm for me. self-test was unaffected too.

Issue #713, opened 2017-01-05
capture modules-db results in ERROR: 757: unexpected token


joeg13:
I'm receiving an ERROR: 757: unexpected token when trying to capture a modules database.  To replicate this, create a modules database on localhost named test-modules and insert a single module named /foo.xqy.  Create a new Roxy project and run the capture command from within the project.  I've tried this with both the master and dev branches.

$ ml local capture --modules-db=test-modules
Retrieving source from test-modules...
ERROR: 757: unexpected token at '--0e040f1aa62b4e54
Content-Type: text/plain
X-Primitive: string

/foo.xqy
--0e040f1aa62b4e54--
'
ERROR: ["/Users/jglorioso/Downloads/temp/deploy/lib/json/common.rb:148:in `parse'", "/Users/jglorioso/Downloads/temp/deploy/lib/json/common.rb:148:in `parse'", "/Users/jglorioso/Downloads/temp/deploy/lib/server_config.rb:1621:in `save_files_to_fs'", "/Users/jglorioso/Downloads/temp/deploy/lib/server_config.rb:1432:in `capture'", "deploy/lib/ml.rb:168:in `<main>'"]

dmcassel:
What does foo.xqy have in it? Did you insert it using "ml deploy modules" or some other way?

joeg13:
In this case, I just used xdmp:document-insert in qconsole.  

```
xquery version "1.0-ml";

let $module :=
document {
'declare namespace html = "http://www.w3.org/1999/xhtml";
<p>hello world</p>'
}
return xdmp:document-insert('/foo.xqy', $module)
```

I have tried this on a client's live Modules database and my local Modules database also.  In each case, I received the same error.  The foo.xqy was just an easily reproducible case for you to test.  

I don't have another readily available environment to test in.  I don't think it matters, but I'm on Mac running OSX 10.12.  As I mentioned, I tried both the master and dev branches of Roxy so I know that it is current.

dmcassel:
Forgot to ask: what version of MarkLogic? With that information, we should have enough to reproduce.

joeg13:
I'm sorry.  That should have been on my initial post.  I've tried this against Linux 8.0-6 and Mac 8.0-6.

dmcassel:
verified in 8.0-5.8 (Mac). Investigating.

dmcassel:
[...]

PR #715, opened 2017-01-13
#713 Correct capture processing to handle multi-part responses

dmcassel:


dmcassel:
One thing that isn't perfect: some captured files will be different from the original in having or not having a final newline.

RobertSzkutak:
First, this PR is to master not dev. I imagine that was an accident?

Second, I have fixed the newline issue. Don't know if its the "best" fix or not though. Basically, there is an extra "\r\n" attached to the payload and I just remove the last "\r\n". A little bit of reading suggests that this may not actually be in the payload but rather that Ruby may be adding it artificially. Either way, its there in the payload and I just remove it at the very end before saving to the filesystem.

You can view my fix here: https://github.com/RobertSzkutak/roxy/commit/2bb98174684fe0ab20209180ca50e629ec6c378e

I was going to send this fix to you as a PR, but the commit history looked very messy since this was made on top of master rather than on top of dev. I'll follow your lead to resolve.

PR #717, opened 2017-01-16
#713 Correct capture processing to handle multi-part responses

dmcassel:


dmcassel:
@RobertSzkutak good catch on the PR to master; thank you. 

I added your chomp line and that's enough to make everything come through correctly. If I make other change you suggested, I get some lines that get collapsed into one. 

My test procedure:

1. init an --app-type=mvc project in my git/roxy-dmc directory
2. `ml local bootstrap`
3. `ml local deploy modules`
4. ` ml local capture --modules-db=roxy-modules`
5. use git to compare

src/app/config/config.xqy shows a differences as it should, and src/Deploy gets populated. I think this is now working properly.

RobertSzkutak:
@dmcassel With my other changes I didn't see a difference but I was checking with a text editor and not git diff. On further, thorough testing and checking with more diff tools I concur with your findings.

This is fine for me so I've merged it.

dmcassel:
thanks for the review, @RobertSzkutak

Issue #687, opened 2016-11-08
Set up issue template


dmcassel:
See [https://github.com/blog/2111-issue-and-pull-request-templates](https://github.com/blog/2111-issue-and-pull-request-templates). 

What do we want to request when someone files an issue? 

Here's [a sample from the Data Hub project](https://github.com/marklogic/marklogic-data-hub/blob/1.0-master/.github/ISSUE_TEMPLATE.md).

RobertSzkutak:
Feel free to edit this comment.

Working proposal for bugs:
## Thank you for submitting an issue!

If this is a bug, please use the bug template below. If this is an enhancement request, please use the RFE template below. Before submitting your issue, please do us a favor and search currently open issues. If your issue already exists, please make a comment there advocating your issue be addressed sooner.

###Bug Template

#### The issue

Short description of the problem:

What are the steps to reproduce the problem?

1. One
2. Two
3. Three

#### Tech Specs

Which Operating System are you using?

Which version of MarkLogic are you using?

Which version of Roxy are you using?

###RFE Template

#### The RFE

Short description of the problem:

Your business use case. How much time would this save you? Can you currently work around this missing feature?

Timeframe : How urgently do you feel you need this RFE addressed? Why?

dmcassel:
This works for me.

dmcassel:
merged into dev. I expect it won't show up after hitting "New Issue" until it gets merged to master.

dmcassel:
fixed in dev

PR #718, opened 2017-01-16
#687 adding issue template

dmcassel:


RobertSzkutak:
Looks like what we talked about in the issue. No suggestions from me

PR #719, opened 2017-01-17
#492 verify parameters during setup

dmcassel:
This ended up being bigger than I expected, but I think the changes are all things we've discussed before. Wherever we were previously relying on having the correct structure in ml-config.xml, the code now calls the appropriate admin function to build a proper structure. 

While I was at it, I replaced several try/catch and xdmp:evals with xdmp:value, as @RobertSzkutak and I discussed previously.

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -2477,7 +2477,32 @@ declare function setup:add-geospatial-element-attribute-pair-indexes(
   admin:database-add-geospatial-element-attribute-pair-index(
     setup:remove-existing-geospatial-element-attribute-pair-indexes($admin-config, $database),
     $database,
-    $db-config/db:geospatial-element-attribute-pair-indexes/db:geospatial-element-attribute-pair-index)
+    for $index in $db-config/db:geospatial-element-attribute-pair-indexes/db:geospatial-element-attribute-pair-index
+    return
+      if (setup:at-least-version("6.0-0")) then
+        admin:database-geospatial-element-attribute-pair-index(
+          $index/db:parent-namespace-uri,
+          $index/db:parent-localname,
+          $index/db:latitude-namespace-uri,
+          $index/db:latitude-localname,
+          $index/db:longitude-namespace-uri,
+          $index/db:longitude-localname,
+          $index/db:coordinate-system,
+          $index/db:range-value-positions,
+          ($index/db:invalid-values, "ignore")[1]
+        )
+      else
+        admin:database-geospatial-element-attribute-pair-index(
+          $index/db:parent-namespace-uri,
+          $index/db:parent-localname,
+          $index/db:latitude-namespace-uri,
+          $index/db:latitude-localname,
+          $index/db:longitude-namespace-uri,
+          $index/db:longitude-localname,
+          $index/db:coordinate-system,
+          $index/db:range-value-positions
+        )
+  )
 };
 
 declare function setup:validate-geospatial-element-attribute-pair-indexes(
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -1683,59 +1683,29 @@ declare function setup:add-fields(
   $database as xs:unsignedLong,
   $db-config as element(db:database)) as element(configuration)
 {
-  setup:add-fields-R(
+  admin:database-add-field(
     setup:remove-existing-fields($admin-config, $database),
     $database,
-    $db-config/db:fields/db:field[db:field-name and fn:not(db:field-name = "")]
-  )
-};
-
-declare function setup:add-fields-R(
-  $admin-config as element(configuration),
-  $database as xs:unsignedLong,
-  $field-configs as element(db:field)*) as element(configuration)
-{
-  if ($field-configs) then
-    let $field := $field-configs[1]
+    for $field in $db-config/db:fields/db:field[db:field-name and fn:not(db:field-name = "")]
     return
-    setup:add-fields-R(
       if ($field/db:field-path) then
         if (setup:at-least-version("7.0-1")) then
-          xdmp:eval(
-            'import module namespace admin = "http://marklogic.com/xdmp/admin" at "/MarkLogic/admin.xqy";
-             declare namespace db = "http://marklogic.com/xdmp/database";
-             declare variable $admin-config external;
-             declare variable $database external;
-             declare variable $field external;
-             admin:database-add-field(
-              $admin-config,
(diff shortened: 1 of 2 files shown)


dmcassel:
Also removed recursion in some cases. Some of the admin functions take a sequence of things to add, and instead of passing in a sequence, we were calling the function multiple times. I simplified the code to just FLWOR over the inputs and remove the recursion, where it made sense to do so.

deploy/lib/xquery/setup.xqy
```diff
@@ -2391,18 +2330,27 @@ declare function setup:add-range-field-indexes-helper(
...
+            ($index/db:collation/fn:string(), "")[1], (: ML6 requires xs:string; later requires xs:string? :)
+            $index/db:range-value-positions,
+            $index/db:invalid-values
+          )
+        else
+          admin:database-range-field-index(
+            $index/db:scalar-type,
+            $index/db:field-name,
+            $index/db:collation,
+            $index/db:range-value-positions
```
RobertSzkutak: Should this also include $index/db:invalid-values in ML versions after 6?

deploy/lib/xquery/setup.xqy
```diff
@@ -2417,41 +2365,42 @@ declare function setup:remove-existing-geospatial-element-indexes(
...
-      else
-        setup:validation-fail(fn:concat("Database mismatched range field index: ", $expected/db:field-name))
-  }
-  catch($ex)
-  {
-    if ($ex/error:code = "XDMP-UNDFUN") then $admin-config
+  let $existing := admin:database-get-range-field-indexes($admin-config, $database)
+  for $expected in $db-config/db:range-field-indexes/db:range-field-index
+  return
+    if ($existing[fn:deep-equal(., $expected)]) then ()
```
RobertSzkutak: Should we also be turning $expected into admin:database-range-field-index ? deep-equal is pretty strict and I would be concerned of formatting/typo issues in ml-config which Roxy ignored but could cause it to fail here.

deploy/lib/xquery/setup.xqy
```diff
@@ -2417,41 +2365,42 @@ declare function setup:remove-existing-geospatial-element-indexes(
...
+          ($index/db:point-format, "point")[1],
+          ($index/db:invalid-values, "ignore")[1]
+        )
+      else
+        admin:database-geospatial-element-index(
+          $index/db:namespace-uri,
+          $index/db:localname,
+          $index/db:coordinate-system,
+          $index/db:range-value-positions,
+          ($index/db:point-format, "point")[1]
```
RobertSzkutak: Should invalid-values be here?

deploy/lib/xquery/setup.xqy
```diff
@@ -2483,7 +2432,32 @@ declare function setup:add-geospatial-element-attribute-pair-indexes(
...
+      else
+        admin:database-geospatial-element-attribute-pair-index(
+          $index/db:parent-namespace-uri,
+          $index/db:parent-localname,
+          $index/db:latitude-namespace-uri,
+          $index/db:latitude-localname,
+          $index/db:longitude-namespace-uri,
+          $index/db:longitude-localname,
+          $index/db:coordinate-system,
+          $index/db:range-value-positions
```
RobertSzkutak: Should invalid values be here?

deploy/lib/xquery/setup.xqy
```diff
@@ -2515,7 +2489,32 @@ declare function setup:add-geospatial-element-pair-indexes(
...
+      else
+        admin:database-geospatial-element-pair-index(
+          $index/db:parent-namespace-uri,
+          $index/db:parent-localname,
+          $index/db:latitude-namespace-uri,
+          $index/db:latitude-localname,
+          $index/db:longitude-namespace-uri,
+          $index/db:longitude-localname,
+          $index/db:coordinate-system,
+          $index/db:range-value-positions
```
RobertSzkutak: Should invalid values be here?

deploy/lib/xquery/setup.xqy
```diff
@@ -2547,7 +2546,30 @@ declare function setup:add-geospatial-element-child-indexes(
...
+        )
+      else
+        admin:database-geospatial-element-child-index(
+          $index/db:parent-namespace-uri,
+          $index/db:parent-localname,
+          $index/db:namespace-uri,
+          $index/db:localname,
+          $index/db:coordinate-system,
+          $index/db:range-value-positions,
+          ($index/db:point-format, "point")[1]
```
RobertSzkutak: Should invalid values be here?

RobertSzkutak:
Outside of the (minor) comments I made this looks very good and appears to work as expected.

deploy/lib/xquery/setup.xqy
dmcassel: it does ([line 2344](https://github.com/marklogic/roxy/pull/719/files#diff-a5c7978e290f869b02ae3656481e20c1R2344)); the `if` section is versions 6+.

RobertSzkutak: @dmcassel Wow, I totally misread that one. Deleted a lot of my other comments. Thanks.

deploy/lib/xquery/setup.xqy
dmcassel: I was thinking that since this is for validation, not creating indexes, it would be okay, but the fn:deep-equal would be thrown off by comments or ordering. Good catch, I'll change this.

dmcassel: I like that calling admin:database-range-field-indexes protects from comments and ordering differences, but I see a different problem. Consider this from ml-config:

        <range-field-index>
          <scalar-type>dateTime</scalar-type>
          <field-name>field-test</field-name>
          <range-value-positions>false</range-value-positions>
          <invalid>reject</invalid>
        </range-field-index>

I've incorrectly specified "invalid" instead of "invalid-values" as the element name. The validate code doesn't catch it, but fn:deep-equal would. Likewise, in any case where the value is optional or an empty string is okay, constructing an element won't work for the comparison, because it will allow the same mistake that the index-creating code makes. 

I've also noticed that the validate code isn't actually getting called. `setup:validate-install` calls `setup:validate-databases-indexes`, which calls `setup:validate-range-field-indexes`, but nothing calls `setup:validate-install` (seems like `setup:do-setup` should). 

In light of the first observation, I propose we keep the comparisons as they are. Based on the second, I suggest we open another issue to address that.

RobertSzkutak: Agreed. I'm okay with merging this as-is

Issue #733, opened 2017-02-04
Bootstrap of range-field-index fails after upgrade to 1.7.5 (due to missing invalid-values in ml-config)


grtjn:
With a range-field-index like:

```
        <range-field-index>
          <scalar-type>string</scalar-type>
          <field-name>words</field-name>
          <collation>http://marklogic.com/collation/</collation>
          <range-value-positions>false</range-value-positions>
        </range-field-index>
```

bootstrap fails with:

```
2017-02-04 11:36:29.155 Info: App-Services: <error:error xsi:schemaLocation="http://marklogic.com/xdmp/error error.xsd" xmlns:error="http://marklogic.com/xdmp/error" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
2017-02-04 11:36:29.155 Info: App-Services:   <error:code>XDMP-AS</error:code>
2017-02-04 11:36:29.155 Info: App-Services:   <error:name>err:XPTY0004</error:name>
2017-02-04 11:36:29.155 Info: App-Services:   <error:xquery-version>1.0-ml</error:xquery-version>
2017-02-04 11:36:29.155 Info: App-Services:   <error:message>Invalid coercion</error:message>
2017-02-04 11:36:29.155 Info: App-Services:   <error:format-string>XDMP-AS: (err:XPTY0004) $invalid-values as xs:string -- Invalid coercion: () as xs:string</error:format-string>
2017-02-04 11:36:29.155 Info: App-Services:   <error:retryable>false</error:retryable>
2017-02-04 11:36:29.155 Info: App-Services:   <error:expr>$invalid-values as xs:string</error:expr>
2017-02-04 11:36:29.155 Info: App-Services:   <error:data>
2017-02-04 11:36:29.155 Info: App-Services:     <error:datum>()</error:datum>
2017-02-04 11:36:29.155 Info: App-Services:     <error:datum>xs:string</error:datum>
2017-02-04 11:36:29.155 Info: App-Services:   </error:data>
2017-02-04 11:36:29.155 Info: App-Services:   <error:stack>
2017-02-04 11:36:29.155 Info: App-Services:     <error:frame>
2017-02-04 11:36:29.155 Info: App-Services:       <error:uri>/eval</error:uri>
2017-02-04 11:36:29.155 Info: App-Services:       <error:line>2687</error:line>
2017-02-04 11:36:29.155 Info: App-Services:       <error:column>10</error:column>
[...]

PR #746, opened 2017-02-14
#733 corrected invalid-values default; addressed missed case

dmcassel:


RobertSzkutak:
I looked through very carefully and couldn't spot any other missing cases

grtjn:
This does change the default behavior of Roxy for a few cases. Something to keep in mind if someone upgrades Roxy, and bootstraps with an old ml-config, like Venu..

PR #760, opened 2017-04-20
Fix #665

### LINKED ISSUE #665: Configurable /app/config/config.xqy file
Currently it's not possible to include roxy environment variables into SJS configuration modules.
I will be nice to be able to configure which file we intend to use as a central configuration file.

For example:
    application-conf-file=/app/config/config.sjs

Or even better:
    application-conf-file=('/app/config/service1-related-config.xqy', '/app/config/service2-related-config.xqy')

divino:
Note:

Configuration can be loaded using

./ml <env> deploy modules

or

./ml <env> deploy config


deploy/lib/server_config.rb
```diff
@@ -1046,6 +1046,8 @@ def deploy
         deploy_triggers
       when 'rest-config'
         deploy_rest_config
+      when 'config'
+        deploy_config
```
dmcassel: By adding these lines, you're making "ml {env} deploy config" a command line option. I see this more as a modification to the existing deploy_src functionality than a new deploy option. (If we were going to make it an option, it would need to be added to the Help as well.) 

I think this should be dropped.

deploy/lib/server_config.rb
```diff
@@ -1971,6 +1973,54 @@ def deploy_src
...
+                                         :add_prefix => File.join(@properties["ml.modules-root"], "app/config/"),
+                                         :permissions => src_permissions
+        end
+        logger.info "Done deploying application configurations"
+      end
+
+      logger.info "\nLoaded #{total_count} #{pluralize(total_count, "document", "documents")} from #{xquery_dir} to #{xcc.hostname}:#{xcc.port}/#{dest_db} at #{DateTime.now.strftime('%m/%d/%Y %I:%M:%S %P')}\n"
+    end
+  end
+
```
dmcassel: This function repeats a lot of code from deploy_src. Why isn't this just a modification to lines [1934-1936](https://github.com/marklogic/roxy/blob/dev/deploy/lib/server_config.rb#L1934-L1936) and [1945-1947](https://github.com/marklogic/roxy/blob/dev/deploy/lib/server_config.rb#L1945-L1947)? 

The existing deploy_src function still has the same functionality as before, which means duplicated processing. I'd either drop or reduce this function and modify the code in deploy_src, perhaps to call a smaller version of this function.

dmcassel:
This PR creates a file src/app/config/custom-config.xqy, which would be the default value for the new property. I like that you've provided a simple example for this. However, the Roxy Deployer supports a few application types and we need to consider how this would play with each:

- --app-type=rest: after running `ml new` with this app type, no source code is provided (src/* gets deleted)
- --app-type=mvc: Roxy MVC expects `/app/config/config.xqy' to exist and to get substitutions made. I think for an MVC app, the default property value should either point to that file or both config.xqy and custom-config.xqy
- --app-type=bare: no code is provided for bare apps (src/* gets deleted), so the sample file wouldn't exist

My suggestion: 

- move custom-config.xqy to the deploy/sample directory
- update the note in the .properties documentation to point to it
- ensure that for MVC apps, the property gets set to either `application-conf-file=src/app/config/custom-config.xqy,src/app/config/config.xqy` or `application-conf-file=src/app/config/config.xqy`

dmcassel:
In my review, I suggested dropping the `deploy_config` functionality. I could be persuaded otherwise. @grtjn, @RobertSzkutak, what do you think?

RobertSzkutak:
@dmcassel I concur with not copy-pasting functions and instead relying on existing functions to deploy this file. I would suggest sticking with `deploy_src` (since the config file is already deployed from there...) and in a future PR enabling a way to select which files to deploy from `deploy_src`. I also agree with your suggestion to move the configuration file to avoid unintended wiping.

divino:
Implemented review point. Kindly review again. Thank you.

Dino Bagayaua
Senior Field Engineer
MarkLogic Corporation
Dino.Bagayaua@marklogic.com
Cell: +63-998-569-5332
www.marklogic.com

dmcassel:
This is better. However, I think I didn't do a good job explaining one of my suggested changes. As is, you have the custom-config.xqy file in the deploy/sample directory, as I requested. The step I left out is that at project initialization, the file should be copied into the src directory (into `src/config/`). Take a look at the `self.init` function and see how for `--app-type=rest` projects, [options and properties files get copied from the sample directory to the appropriate place](https://github.com/marklogic/roxy/blob/dev/deploy/lib/server_config.rb#L270) (under rest-api). Similarly, if this project is `--app-type=rest` or `--app-type=bare`, I would copy `deploy/sample/custom-config.xqy` to `src/config/config.xqy`. For `--app-type=mvc`, just stick with the defaults. With that approach, the code will be in the right place under the `src` directory and you won't have to mess with the properties (just keep the default property as you have it). 

The reason we don't want to just leave it under deploy/sample is that files there would be overwritten during an `ml upgrade`. By copying it to src, the developer's changes will be safe.

PR #761, opened 2017-04-20
Fix #703

### LINKED ISSUE #703: Timestamp-verified restart
A restart function that verifies the restart was successful. Convenient for build scripts. Based on [this example](https://docs.marklogic.com/guide/admin-api/cluster#id_38205).

```
  def verified_restart
    old_timestamp = go(%Q{http://#{@properties["ml.server"]}:8001/admin/v1/timestamp}, "get").body
    restart
    retry_count = 0
    retry_max = 5
    retry_interval= 10
    new_timestamp = old_timestamp
    while retry_count < retry_max do
      begin
        new_timestamp = go(%Q{http://#{@properties["ml.server"]}:8001/admin/v1/timestamp}, "get").body
      rescue
        logger.debug 'Retry attempt ' + retry_count.to_s + ' failed'
      end
      if new_timestamp != old_timestamp
        logger.info 'Verfied restart.'
        break
      end
      logger.debug "Verifying restart..."
      sleep retry_interval
      retry_count += 1
    end
    if new_timestamp == old_timestamp
      logger.warn "Could not verify restart"
    end
  end
end
```

divino:
Note:

We can still skip verification by passing --verify=false

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -566,8 +566,8 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
       old_timestamp = go(%Q{http://#{@properties["ml.server"]}:8001/admin/v1/timestamp}, "get").body
       restart_basic
       retry_count = 0
-      retry_max = 5
-      retry_interval= 10
+      retry_max = @properties["ml.verify_retry_max"].to_i
+      retry_interval = @properties["ml.verify_retry_interval"].to_i
       new_timestamp = old_timestamp
       while retry_count < retry_max do
         begin
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -520,7 +520,7 @@ but --no-prompt parameter prevents prompting for password. Assuming 8.'
     return r
   end
 
-  def restart
+  def restart_basic
     @ml_username = @properties['ml.bootstrap-user'] || @properties['ml.user']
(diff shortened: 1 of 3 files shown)


deploy/lib/server_config.rb
```diff
@@ -556,6 +556,43 @@ def restart
...
   end
 
+  # implemented verified restart
+  def restart
+    verify = find_arg(['--verify'])
+    if verify==='false'
+      restart_basic
+    else
+      # defaults to verified restart
+      old_timestamp = go(%Q{http://#{@properties["ml.server"]}:8001/admin/v1/timestamp}, "get").body
```
dmcassel: didn't know this endpoint existed. nice!

deploy/lib/server_config.rb
```diff
@@ -556,6 +556,43 @@ def restart
...
+      restart_basic
+      retry_count = 0
+      retry_max = 5
+      retry_interval= 10
```
dmcassel: I think at least one of these should be property-driven -- desired values may differ depending on the environment. Please add a property to default.properties and make use of it here.

divino:
implemented review points

divino:
Implemented review points. Kindly review again. Thank you.

Dino Bagayaua
Senior Field Engineer
MarkLogic Corporation
Dino.Bagayaua@marklogic.com
Cell: +63-998-569-5332
www.marklogic.com

PR #767, opened 2017-05-26
Fixed #753: added new use-https-for-rest property

### LINKED ISSUE #753: use-https should control use of https for REST configuration
#### Short description of the problem:

Even with the "use-https=true" property set, REST configuration does not use https unless the "ml.ssl-certificate-template" property is set to something (it can be set to anything).

This is checked here: https://github.com/marklogic/roxy/blob/master/deploy/lib/server_config.rb#L2173

This is mainly relevant where a proxy or load balancer sits in front of MarkLogic and uses https to clients but http to MarkLogic. Because there is no ssl configuration on the MarkLogic side, "ml.ssl-certificate-template" would not be set.

Shouldn't the REST use https if the "use-https=true" property is set?

#### What are the steps to reproduce the problem?

In the deploy/build.properties
1. Set "use-https=true"
2. Make sure "ml.ssl-certificate-template" is commented out 
3. Add REST configuration files in the rest-api directory
4. run ./ml <env> deploy modules

#### Tech Specs

Which Operating System are you using?
Linux

Which version of MarkLogic are you using?
8.0-6.3

Which version of Roxy are you using (see version.txt)?
1.7.3+

grtjn:
Fixes #753

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2236,7 +2236,7 @@ private
         :http_connection_open_timeout => @properties["ml.http.open-timeout"].to_i,
         :http_connection_read_timeout => @properties["ml.http.read-timeout"].to_i,
         :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i,
-        :use_https_for_rest => @properties["ml.ssl-certificate-template"].present?
+        :use_https_for_rest => @properties["ml.ssl-certificate-template"].present? || @properties["ml.use-https-for-rest"] == "true"
       })
     else
       @mlRest
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -193,6 +193,12 @@ install-xcc=true
 #
 use-https=false
 
+#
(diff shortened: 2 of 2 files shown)


deploy/lib/server_config.rb
```diff
@@ -2236,7 +2236,7 @@ def mlRest
         :http_connection_open_timeout => @properties["ml.http.open-timeout"].to_i,
         :http_connection_read_timeout => @properties["ml.http.read-timeout"].to_i,
         :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i,
-        :use_https_for_rest => @properties["ml.ssl-certificate-template"].present?
+        :use_https_for_rest => @properties["ml.ssl-certificate-template"].present? or @properties["ml.use-https-for-rest"] == "true"
```
grtjn: Typo.. `or` should be `||`

RobertSzkutak:
Pretty simple change. Looks good to me.

PR #768, opened 2017-05-28
Fixed #661: wrap app_name query in xdmp:eval to provide appropriate context

### LINKED ISSUE #661: Allow execute_query with app_name set to ml.app-name

grtjn:
Fixes #661

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2412,8 +2412,10 @@ private
     # 4. Any database
     if properties[:db_name] != nil
       db_id = get_db_id(properties[:db_name])
+      logger.warn "WARN: No Database with name #{properties[:db_name]} found" if db_id.nil?
     elsif properties[:app_name] != nil
       sid = get_sid(properties[:app_name])
+      logger.warn "WARN: No App-Server with name #{properties[:app_name]} found" if sid.nil?
     else
       sid = get_sid("Manage")
     end
@@ -2454,8 +2456,10 @@ private
     # 4. Any database
     if properties[:db_name] != nil
       db_id = get_db_id(properties[:db_name])
+      logger.warn "WARN: No Database with name #{properties[:db_name]} found" if db_id.nil?
     elsif properties[:app_name] != nil
       sid = get_sid(properties[:app_name])
+      logger.warn "WARN: No App-Server with name #{properties[:app_name]} found" if sid.nil?
     else
       sid = get_sid("Manage")
     end
@@ -2487,11 +2491,22 @@ private
   end
 
   def execute_query_8(query, properties = {})
+    # check input like in older versions
+    if properties[:db_name] != nil
+      db_id = get_db_id(properties[:db_name])
+      raise ExitException.new("No Database with name #{properties[:db_name]} found") if db_id.nil?
+    elsif properties[:app_name] != nil
+      sid = get_sid(properties[:app_name])
+      raise ExitException.new("No Server with name #{properties[:app_name]} found") if sid.nil?
+    end
+
     headers = {
       "Content-Type" => "application/x-www-form-urlencoded"
     }
     params = {}
 
+    # If app_name is specified, wrap the eval in an xdmp:eval to create an eval context
+    # that matches that of the selected app-server
     if properties[:app_name] != nil
       params[:xquery] = %Q{
         xquery version "1.0-ml";
@@ -2506,7 +2521,10 @@ private
           let $modules-id := xdmp:server-modules-database($server-id)
           let $xquery-version := xdmp:server-default-xquery-version($server-id)
           let $modules-root := xdmp:server-root($server-id)
-          let $default-coordinate-system := xdmp:server-coordinate-system($server-id)
+          let $default-coordinate-system :=
+            (: xdmp:server-coordinate-system not supported in ML8 and older :)
+            for $f in fn:function-lookup(xs:QName("xdmp:server-coordinate-system"), 1)
+            return $f($server-id)
           return
             <options xmlns="xdmp:eval">{
               if ($database-id eq xdmp:database()) then ()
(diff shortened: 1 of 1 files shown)


grtjn:
In case someone wonders, a simple way to run a quick test is to copy the execute_query_8 into app_specific.rb, and run the test code I provided in this comment of #661: https://github.com/marklogic/roxy/issues/661#issuecomment-281598395

RobertSzkutak:
Having two problems with this PR...

1) On the latest version of ML8, the code throws this error:
ERROR: 500 "Internal Server Error"
ERROR: {"errorResponse":{"statusCode":500, "status":"Internal Server Error", "messageCode":"XDMP-UNDFUN", "message":"XDMP-UNDFUN: (err:XPST0017) Undefined function xdmp:server-coordinate-system()", "messageDetail":{"messageTitle":"Undefined function"}}}

Seems like to fix this its time to finally introduce execute_query_9 ...

2) When I set the db_name param to something other than the app content db (the value used in your second test) on the latest ML8 it throws this error:
ERROR: 500 "Internal Server Error"
ERROR: {"errorResponse":{"statusCode":500, "status":"Internal Server Error", "messageCode":"XDMP-NODB", "message":"XDMP-NODB: xdmp:database-name(xs:unsignedLong(\"0\")) -- No database with identifier 0", "messageDetail":{"messageTitle":"No database"}}}

Too tired to figure this one out I will try to debug it in the morning.

grtjn:
Thanks, good finds. I should have tested my code against other than latest. I'll poke some more at this..

grtjn:
Issue 2 might just be bad test code. The bit of code I referred to above tries to get the database-name, but does not anticipate modules-db being set to (file system), which is the case for no params, since that reverts to the Manage app-server. Use this code instead:

```
  def test
    logger.info "Execute query with no params:"
    r = execute_query(
      %Q{
        xquery version "1.0-ml";

        declare function local:database-name($id) {
          if ($id eq 0) then
            "(file system)"
          else
            xdmp:database-name($id)
        };

        "content-db: " || local:database-name(xdmp:database()),
        "modules-db: " || local:database-name(xdmp:modules-database()),
        "schemas-db: " || local:database-name(xdmp:schema-database()),
        "trigger-db: " || local:database-name(xdmp:triggers-database())
      }
    )
    r.body = parse_body r.body
    logger.info r.body
    logger.info ""

    logger.info "Execute query with db_name param:"
    r = execute_query(
      %Q{
        xquery version "1.0-ml";

        declare function local:database-name($id) {
          if ($id eq 0) then
            "(file system)"
          else
            xdmp:database-name($id)
        };

        "content-db: " || local:database-name(xdmp:database()),
        "modules-db: " || local:database-name(xdmp:modules-database()),
        "schemas-db: " || local:database-name(xdmp:schema-database()),
        "trigger-db: " || local:database-name(xdmp:triggers-database())
      },
      { :db_name => @properties['ml.content-db'] }
    )
    r.body = parse_body r.body
    logger.info r.body
    logger.info ""

    logger.info "Execute query with app_name param:"
    r = execute_query(
      %Q{
        xquery version "1.0-ml";

        declare function local:database-name($id) {
          if ($id eq 0) then
            "(file system)"
          else
            xdmp:database-name($id)
        };

        "content-db: " || local:database-name(xdmp:database()),
        "modules-db: " || local:database-name(xdmp:modules-database()),
        "schemas-db: " || local:database-name(xdmp:schema-database()),
        "trigger-db: " || local:database-name(xdmp:triggers-database())
      },
      { :app_name => @properties['ml.app-name'] }
    )
    r.body = parse_body r.body
    logger.info r.body
    logger.info ""
  end
```

Also make sure you bootstrap before you run that test, otherwise mentioned values will be wrong (details for Manage or some random app-server).

I'm investigating running exec_8 against ML8..

grtjn:
This should work better..

I'm printing warnings for ml7- if a referenced db_name or app_name is not found. The execute_query_5/7 seemed to be very forgiving, a bit too much I think.

The execute_query_8 however is not forgiving, no fallback if db_name or app_name refers to non-existing items. I added a check upfront, and raise an exception if non-existing. The effect is similar if we don't do this, but the feedback for the user is cleaner this way.

I also applied a trick with fn:function-lookup to be server-version agnostic. All seems to run smooth against ml7, ml8, and ml9 now..

grtjn:
@RobertSzkutak Can you give it another spin?

RobertSzkutak:
Everything looks good!

PR #769, opened 2017-05-28
Fixed #650: make sure server-version gets defaulted properly in all cases

### LINKED ISSUE #650: Rewriter not initialized correctly when not selecting server-version upfront
If you run ml new and forget to specify server-version, Roxy will ask, but if you pick 8, it still initializes with the wrong xqy rewriter. Luckily it auto-corrects at deploy, but it does print a warning.

Minor bug..

grtjn:
Fixes #650

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -323,15 +323,19 @@ class ServerConfig < MLClient
   def self.prompt_server_version
     if @@no_prompt
       puts 'Required option --server-version=[version] not specified with valid value,
-but --no-prompt parameter prevents prompting for password. Assuming 8.'
+but --no-prompt parameter prevents prompting for password.'
+      server_version = 0
     else
       puts 'Required option --server-version=[version] not specified with valid value.
 
   What is the version number of the target MarkLogic server? [5, 6, 7, 8, or 9]'
       server_version = STDIN.gets.chomp.to_i
-      server_version = 8 if server_version == 0
-      server_version
     end
+    if server_version == 0
+      puts "Defaulting to 8.."
+      server_version = 8
+    end
+    server_version
   end
 
   def self.index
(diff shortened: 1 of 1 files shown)


grtjn:
There was a merge conflict, but I resolved that with git rebase..

grtjn:
To test this, run `./ml new` with and without `--server-version`, with and without `--no-prompt`, and when prompted both try entering a number as well as just hitting enter.

deploy/lib/server_config.rb
```diff
@@ -318,15 +318,19 @@ def self.initcpf
...
       puts 'Required option --server-version=[version] not specified with valid value.
 
   What is the version number of the target MarkLogic server? [5, 6, 7, 8, or 9]'
       server_version = STDIN.gets.chomp.to_i
-      server_version = 8 if server_version == 0
-      server_version
     end
+    if server_version == 0
+      puts "Defaulting to 8.."
+      server_version = 8
```
RobertSzkutak: Would it make more sense to default to 9?

deploy/lib/server_config.rb
```diff
@@ -318,15 +318,19 @@ def self.initcpf
   def self.prompt_server_version
     if @@no_prompt
       puts 'Required option --server-version=[version] not specified with valid value,
-but --no-prompt parameter prevents prompting for password. Assuming 8.'
+but --no-prompt parameter prevents prompting for password.'
+      server_version = 0
     else
       puts 'Required option --server-version=[version] not specified with valid value.
 
   What is the version number of the target MarkLogic server? [5, 6, 7, 8, or 9]'
```
RobertSzkutak: Change this to only say 7, 8, or 9?

grtjn:
Makes sense, I'll look into it tomorrow..

grtjn:
What do you think of this? The warning is printed each time a command with an env is used if I am not mistaken. Should we add an option to silence the warning? The one for old Ruby is also always shown I think..

RobertSzkutak:
Looks good to me!

Issue #735, opened 2017-02-09
replicating Security DB results in failed bootstrap


joecrean:
#### The issue

Short description of the problem:

When you try to replicate the Security database as part of a bootstrap  you can see that when the Security db goes offline the bootstrap seems to continue and gets into a kind of death loop trying to lookup details in the Security db. Here is a sample of the output in the ML ErrorLog.txt.

```
2017-02-09 14:20:12.583 Debug: Retrying HTTPRequestTask::handleXDBCEval eval 8527616213787243855 Update 90 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:14.487 Debug: Retrying ForestReindexerThread::reindexFragments Last-Login 28 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:14.843 Debug: Retrying HTTPRequestTask::handleXDBCEval eval 8527616213787243855 Update 91 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:15.918 Debug: Retrying ForestReindexerThread::reindexFragments Modules 28 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:16.961 Debug: Retrying HTTPRequestTask::handleXDBCEval eval 8527616213787243855 Update 92 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:16.975 Debug: Retrying ForestReindexerThread::reindexFragments Meters 30 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:17.350 Debug: Retrying ForestReindexerThread::reindexFragments Schemas 29 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:17.819 Debug: Retrying ForestReindexerThread::reindexFragments Fab 25 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:18.728 Debug: Retrying ForestReindexerThread::reindexFragments App-Services 27 because XDMP-FORESTNOT: Forest Security not available: middle closing
2017-02-09 14:20:18.831 Debug: Retrying ForestReindexerThread::reindexFragments Documents 28 because XDMP-FORESTNOT: Forest Security not available: middle closing
[...]

PR #774, opened 2017-06-01
#735: accidental move of create-amps, breaking self-test

grtjn:
Additional fix for #735, should not have moved the create-amps in the first place. This restores it to where it used to be..


grtjn:
I need to merge this, otherwise I can't continue testing CI for self-test on the dev branch..

PR #781, opened 2017-06-01
Fixed #749: added app-modules-db prop for legacy with dep warning

### LINKED ISSUE #749: Upgrading Roxy in projects still using app-modules-db causes bootstrap to fail
app-modules-db references were removed from the project, but if you upgrade an existing project they are likely still present. Roxy will complain at bootstrap that the ml-config.xml refers to an unknown property, with a not too helpful error message like this:

```
Bootstrapping your project into MarkLogic 9 on ea4-ml1...
ERROR: note: restart required
XDMP-NOSUCHDB: xdmp:database("@ml.app-modules-db") -- No such database @ml.app-modules-db
See MarkLogic Server error log for more details.
ERROR: ... Bootstrap FAILED
```

I recommend reinstating it in default.properties, defaulting to $modules-db, just for the purpose of legacy support..

grtjn:
Fixes #749 

This will make a lot of people happy when they try to upgrade older projects..

### DIFF
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -108,6 +108,9 @@ app-role=${app-name}-role
 content-db=${app-name}-content
 modules-db=${app-name}-modules
 
+# Note: purely for legacy support!
+app-modules-db=${modules-db}
+
 #
 # Number of forests to create per host in the group for the content-db
 #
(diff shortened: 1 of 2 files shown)


RobertSzkutak:
Tested this and it worked as expected. Pretty simple and much needed fix. My only complaint is that it throws the warning for comments containing app-modules-db. Regardless, merging...

grtjn:
Let's consider that a feature.. :-D

PR #784, opened 2017-06-06
Fixed #762: Made pipeline-config.xml configurable

### LINKED ISSUE #762: CPF deploy config file
#### Parametirized pipeline-config file

We are using for a project with multiple environments. We therefore have multiple $env.properties file.
We are using CPF and we also need to have different pipeline-config files but the "pipeline-config.xml" is hard coded and this is not possible.

We have made the change in server_config.rb (monkey patching as you described was also an option) by using a property variable instead of the hard coded filename.

Roxy version : 1.7.5

grtjn:
Fixes #762

### DIFF
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -276,3 +276,9 @@ application-conf-file=src/app/config/config.xqy
 #
 verify_retry_max=5
 verify_retry_interval=10
+
+#
+# CPF Pipeline configuration file
+#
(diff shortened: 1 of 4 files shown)


RobertSzkutak:
Hadn't tried CPF before. The code appeared straightforward. Ran several deployments and determined I could change path of pipeline-config.xml successfully. self-test is ok. Merging in...

PR #785, opened 2017-06-07
Fixed #500: linking sample tasks to bootstrap host by default

### LINKED ISSUE #500: Don't add schedules to all hosts by default

grtjn:
Fixes #500

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -104,6 +104,9 @@ class ServerConfig < MLClient
     else
       @qconsole_port = @bootstrap_port
     end
+
+    r = execute_query %Q{xdmp:host-name()}
+    @properties["ml.server-name"] = parse_body(r.body)
   end
 
   def get_properties
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -105,8 +105,12 @@ class ServerConfig < MLClient
       @qconsole_port = @bootstrap_port
     end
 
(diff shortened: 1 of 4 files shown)


deploy/lib/server_config.rb
```diff
@@ -104,6 +104,9 @@ def initialize(options)
     else
       @qconsole_port = @bootstrap_port
     end
+
+    r = execute_query %Q{xdmp:host-name()}
+    @properties["ml.server-name"] = parse_body(r.body)
```
RobertSzkutak: Seems like this would also have the side effect of not allowing people to hit their load balancer? Couldn't that potentially be unwanted behavior for tools like MLCP?

grtjn: Can you elaborate? Keep in mind server-name is only used within ml-config. We will remain using ml.server to make http/xcc calls..

RobertSzkutak: You're correct, I confused the two properties

grtjn:
Yeah, the ml.server-name property is new, and not read from properties yet. I could add that though, as option. And I need to make the xqy call a bit more robust. (capture exceptions if ml cannot be reached or such..)

RobertSzkutak:
This is fine for me. Appeared to work correctly and didnt seem to break anything.

PR #786, opened 2017-06-07
Fixed #531: deploy rest options against test-port

### LINKED ISSUE #531: rest options are not deployed to the test modules database
I'm defining a test-content and test-modules database. When I deploy modules all content from src will be deployed to the test-modules database, but the rest extensions and the rest options are not deployed to the test-modules database. I do have a standard rest api on the test-port but I'm not able to utilize the rest extensions and options on the test-port.
Is that by design? Can they be added?

grtjn:
Fixes #531. Not entirely convinced this is the best way, but clean and simple. You need to use deploy rest or deploy modules to deploy rest artifacts for testing, not deploy rest-config, transforms or ext. This felt sufficient for now though..

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2039,7 +2039,7 @@ private
     end
   end
 
-  def deploy_rest
+  def deploy_rest(test = false)
     # Deploy options, extensions to the REST API server
     if ['rest', 'hybrid'].include? @properties["ml.app-type"]
       # Verify that we're not trying to run REST from the filesystem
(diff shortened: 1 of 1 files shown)


deploy/lib/server_config.rb
```diff
@@ -2058,6 +2058,37 @@ def deploy_rest
       deploy_rest_config()
       deploy_ext()
       deploy_transform()
+
+      if !test &&
```
RobertSzkutak: Whats the point of `test` ? Would it only be true if called by app_specific.rb ?

grtjn: No, deploy_rest calls itself to repeat the deploy stuff against test-port. See following lines, I create a new temporary mlRest client, targeting test port and such, and then make a recursive call with a true flag to stop further recursion..

RobertSzkutak: I see. Thanks!

PR #787, opened 2017-06-07
Fixed #603: consume --ml options rather than breaking on them

### LINKED ISSUE #603: Command-line overrides for props are not removed from ARGV
util.rb, load_prop_from_args should remove args matching the --ml.xx= pattern. I was able to fix that by copying these two lines from the find_arg function, and adding those after the `if a.match...`:

``` ruby
      index = ARGV.index(a)
      ARGV.slice!(index)
```

grtjn:
Fixes #603


RobertSzkutak:
Straightforward fix

PR #790, opened 2017-06-07
Fixed #622: applied escaping of props only for xquery use

### LINKED ISSUE #622: Escape passwords on use, not upfront

grtjn:
Fixes #622.

Lines 2716 upto 2854 have only been indented 2 more space characters.

### DIFF
--- a/deploy/lib/MLClient.rb
+++ b/deploy/lib/MLClient.rb
@@ -28,7 +28,7 @@ class MLClient
 
   def initialize(options)
     @ml_username = options[:user_name]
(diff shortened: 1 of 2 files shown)


grtjn:
Use https://github.com/marklogic/roxy/pull/790/files?w=1 for a cleaner view of the changes..

grtjn:
This PR also included a fix for #823..

PR #791, opened 2017-06-07
Quick fix to make self-test work on higher versions of ML8

RobertSzkutak:
This is not an 'ideal' fix but it seems to support 7 and 8 cleanly for now.


grtjn:
I think I'd prefer to start copying ml7 to ml8 and ml9 instead. As noted offline, you can also influence the server-version by just dropping in a local.properties into deploy/test/data/ml7-properties/..

RobertSzkutak:
That's fine

PR #792, opened 2017-06-07
Fixed #664: capture and pass through exit code of mlcp, corb, etc

### LINKED ISSUE #664: MLCP call doesn't return failed code if Java not found
system() returns a boolean, but the result is not checked..

grtjn:
Fixes #664.

The system call invokes sh, which in turn returns a 127 exit code if it cannot find java. The nil case is therefor never reached, but leaving it in place nonetheless. And at least it bails out in case Java is missing.

Side note: `./ml local mlcp version` exits with code 1 too, not sure why..


RobertSzkutak:
Pretty straightforward fix. Forced a success and fail scenario with MLCP and it seemed to log correctly. Didn't test CORB or XQSync but the code looks identical there as it does in MLCP handling.

PR #793, opened 2017-06-08
Fixed #757: return http 409 if failed, and use for exit code

### LINKED ISSUE #757: Returning exit status for unit test invocation
#### Exit status always 0 for unit tests

Currently running unit tests from the terminal only prints the test results to the terminal [source link](https://github.com/marklogic/roxy/blob/0cd9666ae7e18de47da802c5bed02e077143891b/deploy/lib/server_config.rb#L1044). If there are any failing unit tests, the user needs to examine the printed xml for any errors. The exit status will always remain 0 (succeeded).

For example running

``` bash
./ml local test
```
will only return an exit status of 0 despite whether a unit test failed or not.

#### Use case

Returning an exit status for the unit tests would be valuable to automated builds of a project so that a build will fail a unit test fails.

#### Potential solution

One potential way to return the exit status based on the success of unit tests would be to parse the XML output from the test script, /test/default.xqy, and check if the "errors" or "failures" attributes in the testsuite element is anything but 0. 

I plan on writing a solution for my local project. Let me know if you would like me to submit a pull request.

grtjn:
Fixes #757

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -1297,12 +1297,25 @@ In order to proceed please type: #{expected_response}
         suites << line.gsub(/.*suite path="([^"]+)".*/, '\1').strip if line.match("suite path")
       end
 
+      success = true
       suites.each do |suite|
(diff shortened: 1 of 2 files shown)


RobertSzkutak:
Seems straightforward. Got a 409 by running curl on a failing test suite.

PR #798, opened 2017-06-08
Fixes #198: added check of server-version against xdmp:version

### LINKED ISSUE #198: Auto-detect ML version..
Not entirely sure roxy can do detection of ML version fully automatically, but surely it should be able to give a helping hand, and warn when things look out of sync.

Could be as simple as an ml version command that does an educated guess on all environments, and help you push that into the config files..

grtjn:
Fixes #198. Limited implementation, but should be sufficient..


grtjn:
I resolved the merge conflict on this one..

PR #801, opened 2017-06-09
Fixed #799: allow non-local to be wiped without confirmation

### LINKED ISSUE #799: Allow non-confirmed wipe of env's other than local
local is currently the only, and hard-coded environment for which no warning is issues at wipe. But people are allowed to add other, custom environments, which they might wanna give the same treatment.

grtjn:
Fixes #799


RobertSzkutak:
Tested a couple envs and it worked. Straightforward implementation

PR #803, opened 2017-06-12
Fixed #802: check both for CHANGELOG.md and .mdown at up/downgrade

### LINKED ISSUE #802: Upgrade/downgrade between 1.7.0 and 1.7.1 fails
CHANGELOG.md was renamed to CHANGELOG.mdown between 1.7.0 and 1.7.1. This causes upgrader to fail if you upgrade/downgrade between that.

grtjn:
Fixes #802


RobertSzkutak:
Tested and it worked. Small, simple change

PR #817, opened 2017-06-15
Fixed #692: leveraging search:parse to evaluate if's against properties

### LINKED ISSUE #692: Define environment specific assets in ml-config.xml
Request from a customer : 

Have an optional attribute in Roxy assets which defines environments that are allowed to use that asset.

ex:

This asset would only be deployed/wiped on dev environment:

    <role @environment='dev'/>

This asset would only be deployed/wiped on local and dev environment:

    <user @environment='local,dev'>

While you can do this with using a different ml-config,xml for each environment, it would be much more condensed to manage assets like this within a single ml-config.xml file.

grtjn:
Fixes #692.

Adds support to use an if attribute with a boolean expression on any config element, even the root (configuration). It allows expressions like:
- prop EQ val
- prop NE val
- prop LT val
- prop LE val
- prop GT val
- prop GE val
- -(prop EQ val)
- prop1 EQ val1 OR prop1 EQ val2
- prop1 EQ val1 AND prop2 EQ val2
- prop1 EQ val1 AND (prop2 EQ val2 OR prop2 EQ val2)
- prop1 EQ val1 AND -(prop2 EQ val2 OR prop2 EQ val2)
 
Minus represents NOT. For property names one uses the exact names from ml info, but without ml., nor @ or $. E.g.: `if="environment EQ local"`

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -761,12 +761,20 @@ but --no-prompt parameter prevents prompting for password.'
     logger.info r.body
   end
 
+  def properties_map
+    entries = []
+    @properties.each do |k, v|
+      entries.push %Q{map:entry("#{k}", "#{v}")}
+    end
+    "map:new((\n" + entries.join(",\n  ")+ "))"
+  end
+
   def config
     setup = File.read ServerConfig.expand_path("#{@@path}/lib/xquery/setup.xqy")
     r = execute_query %Q{
       #{setup}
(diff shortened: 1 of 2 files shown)


RobertSzkutak:
From my testing I'm very confident the parser doesn't execute without an if attribute. Basic test cases I tried all appeared to work. Feel comfortable merging this. AWESOME feature!

PR #827, opened 2017-08-04
Fix ./ml wipe

RobertSzkutak:
This resolves an invalid coercion error thrown when trying to run ./ml wipe


RobertSzkutak:
Tagging @grtjn so he's aware it was fixed

PR #828, opened 2017-08-04
Fix a bug that breaks bootstraps of projects from ml new

RobertSzkutak:
Create a project with ./ml new

Bootstrap

It fails with ADMIN PORT IN USE

This fixes it. Wasn't caught by self test. Not sure when this broke. There may be a 'better' fix but I'm merging this to urgently fix the issue


RobertSzkutak:
Tagging @grtjn  for awareness

Merging into master and then will merge into DEV from master

PR #829, opened 2017-08-04
Fix issue causing multiple password prompts

RobertSzkutak:
Fix issue causing password to be prompted for multiple times in bootstrap and wipe when it does not exist


RobertSzkutak:
Tagging @grtjn  for awareness.

Merging into master and then merging into dev from master.

Issue #834, opened 2017-08-16
Support for bootstrapping metadata fields


tdiepenbrock:
#### The RFE

Short description of the problem:

Need support for creating/wiping/etc metadata fields

tdiepenbrock:
I have this implemented, will submit a pull request

grtjn:
Fixed in dev

PR #835, opened 2017-08-16
Implementation of issue #834, support for bootstrapping metadata fields.

tdiepenbrock:
Implementation of this ticket for supporting metadata fields.  Without this support metadata fields fail because there is no include-root element. #834

### DIFF
--- a/deploy/sample/ml-config.sample.xml
+++ b/deploy/sample/ml-config.sample.xml
@@ -198,6 +198,16 @@
       <fields>
         <field>
           <field-name></field-name>
+          <metadata/>
+          <field-value-searches>false</field-value-searches>
+          <field-value-positions>false</field-value-positions>
+          <word-lexicons/>
+          <included-elements/>
+          <excluded-elements/>
+          <tokenizer-overrides/>
+        </field>
+        <field>
+          <field-name></field-name>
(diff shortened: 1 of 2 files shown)


grtjn:
Awesome. Up next bi-temp stuff, and you make my day.. :)

Would be nice to add an example in the sample ml-config, and we would probably like to add a testcase in self-test too, but if you are uncomfortable with doing that, we can take that too..

deploy/sample/ml-config.sample.xml
```diff
@@ -198,6 +198,16 @@
...
         <field>
           <field-name></field-name>
+          <metadata/>
+          <field-value-searches>false</field-value-searches>
+          <field-value-positions>false</field-value-positions>
+          <word-lexicons/>
+          <included-elements/>
+          <excluded-elements/>
+          <tokenizer-overrides/>
+        </field>
```
grtjn: Hmm, interesting. Does a metadata field take tokenizers, lexicons and such? Looks like not. Perhaps remove those elems from the sample?

grtjn:
This PR was replaced by PR #866. Thanks @tdiepenbrock, I just tidied and merged multiple PRs into one, your code changes were helpful!

PR #842, opened 2017-09-05
Fix #585 by @rlouapre

### LINKED ISSUE #585: Support for schema, namespaces, module locations and traces for groups

grtjn:
Fix for #585. Rebased to the latest dev branch.

### DIFF
--- a/deploy/lib/xquery/setup.xqy
+++ b/deploy/lib/xquery/setup.xqy
@@ -140,6 +140,11 @@ declare variable $group-settings :=
     <!--TODO: setting>audit-role-restriction</setting>
     <setting>audit-uri-restriction</setting>
     <setting>audit-user-restriction</setting-->
+    <setting min-version="7.0-0" function="add" keys="namespace-uri location">module-location</setting>
+    <setting min-version="7.0-0" function="add" keys="prefix namespace-uri">namespace</setting>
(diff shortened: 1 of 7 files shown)


grtjn:
Don't merge yet. I'd like to review the xdmp:eval bit, and want to copy over the ml7-config changes to ml8 and ml9 configs too..

grtjn:
This PR replaces #676. For comments on xdmp:eval and xdmp:value see also there..

grtjn:
The code already makes a lot of use of xdmp:value, and since the xdmp:eval is within a validate step (which is essentially only used for self-test), I think we can postpone polishing of that bit..

grtjn:
We could actually remove a lot of evals and values, since ML6 has been deprecated, and ML7 will follow soon too. But perhaps that is better dealt separately when ML7 has been officially deprecated..

grtjn:
I synched the self-test configs, and improved them a little too. Runs clean against ml9. If we merge into dev, CI will self-test against ML8 and ML7.

@RobertSzkutak / @dmcassel Agree to merge?

grtjn:
PS: thanks to @rlouapre for the hard part of the work..

grtjn:
Need to merge this be able to fix merge conflicts with other PR's

Issue #836, opened 2017-08-16
Deploying Source on app-type=bare also deploys the app directory


alan-johnson:
#### The issue

When the the app-type=bare, running "ml <env> deploy modules" or "ml <env> deploy src" throws an error that Roxy can't find app/config/config.xqy in the "src" directory. After adding that file in the required path, it also gets deployed to the modules database. When app-type=bare, it shouldn't throw an error if the app/config/config.xqy is missing since it is not a requirement for the deployment.

What are the steps to reproduce the problem?
1. Create a project with app-type=bare as a build.property
2. Make sure src has a file or two but no "app" directory
3. Run "ml <env> deploy modules"

#### Tech Specs

Which Operating System are you using?
Mac OSX Sierra, 10.12

Which version of MarkLogic are you using?
Version 9.0.1 and Version 8.0.5

Which version of Roxy are you using (see version.txt)?
"July-2017"

grtjn:
Yeah, makes sense. It should probably flush the value of the config property at ml new/init for app-type bare. And it shouldn't error out, but rather warn that the app config is missing. I've seen this cause issues at upgrades too..

grtjn:
Fixed in dev

PR #843, opened 2017-09-05
Fixed #841: no app-config for app-type bare

### LINKED ISSUE #841: Deploy modules fails after upgrade
After upgrade deploy modules returns:

```
ERROR: No such file or directory - src/app/config/config.xqy
```

I think it would be better if `application-conf-file` would have an empty default value.

grtjn:
Fixes #841 and #836

### DIFF
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -276,7 +276,7 @@ http.retry-delay=15
 #
 # Application configuration files
 #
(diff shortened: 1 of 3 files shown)


grtjn:
Annoying issue, small fix. Merging this, so that CI can self-test it..

PR #867, opened 2018-01-10
Fixed #849: slipped around deploy schemas/modules order in install

### LINKED ISSUE #849: Install should deploy schemas before modules
I had a rest extension that was importing a schema, and ./ml local install failed because schemas are deploy after modules. Should be other way round..

grtjn:
#849


grtjn:
Tiny change, merging now..

PR #868, opened 2018-01-10
Fixed #821: allow executing sjs queries

### LINKED ISSUE #821: Allow running JavaScript using execute_query
Would only take passing in :javascript => true in properties, and responding to that. Could be a ML8+ only..

While at it, could also support sparql, by calling /v1/graphs/sparql. ML8+ only though..

grtjn:
#821


grtjn:
New functionality doesn't impact default behavior, merging now..

PR #870, opened 2018-01-10
Fixed #822: added prop replace on mlcp args

### LINKED ISSUE #822: Var expansion in MLCP command-line arguments
Placeholders are replaced inside options files, but not in command-line. One-line change to fix that.

Could be applied to commands like corb and such too..

grtjn:
#822


grtjn:
One-line change, merging now..

PR #871, opened 2018-01-11
Fixed #819: allow skipping of indexes, fragment roots, lexicons, fields, etc. Also some code cleanup.

### LINKED ISSUE #819: Don't flush indexes if index wrapper element is omitted
setup:configure-indexes always flushes indexes, and reconfigures them. It would however give people more control over what happens if flushing them would be skipped if corresponding index wrapper element is omitted (as opposed to just leaving it empty as in the OOTB ml-config.xml).

grtjn:
Fixes #819

It will skip all repeatable artifacts of databases if either the wrapper element (`<element-range-indexes>`, `<fragment-roots>`, etc) are omitted, or it the wrapper is decorated with a `append="true"`.

I added testing for this to self-test, for which I had to make sure all bootstrap test are executed in declared order, so had to bundle them all in one test. It speeds up self-test by the way, as it is no longer doing an extra (unnecessary) bootstrap, ánd wipe, for each of the test cases.

I also took the opportunity to remove some legacy. I unwrapped all xdmp:value's and xdmp:eval's concerning ML6, and straightened out a few unnecessary code loops.

### DIFF
--- a/deploy/lib/MLClient.rb
+++ b/deploy/lib/MLClient.rb
@@ -110,7 +110,7 @@ class MLClient
     end
 
     request_params[:request].use_xcc(xcc)
-    logger.debug(params)
+    #logger.debug(params)
     response = get_http.request request_params
     response.value
     response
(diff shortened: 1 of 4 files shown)


grtjn:
I also wrapped the geo path index stuff to make it run against ML7 and ML8. I'll merge now, so that CI will stop complaining about that..

PR #879, opened 2018-04-05
Fixed #878: added --quiet flag, auto-enabled for info command

### LINKED ISSUE #878: output of info in json format prefixed with WARN messages
Recent changes add warnings in case server cannot be reached, but the WARN messages pollute the JSON output.

grtjn:


grtjn:
This is bothering slush-marklogic-node projects, merging now. I ran local tests in real projects to make sure it runs smooth..

Issue #659, opened 2016-09-09
ml jar broken with jruby >= 9


simonepstein:
If I build the jar having installed a warbler version >= 2 I get the following error when running local deploy

java -jar roxy.jar local bootstrap
LoadError: no such file to load -- app_specific
  require at org/jruby/RubyKernel.java:956
  require at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55
   <main> at uri:classloader:/roxy/bin/roxy.rb:39
     load at org/jruby/RubyKernel.java:974
   <main> at uri:classloader:/META-INF/main.rb:1
  require at org/jruby/RubyKernel.java:956
   (root) at uri:classloader:/META-INF/main.rb:1
   <main> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
ERROR: org.jruby.embed.EvalFailedException: (LoadError) no such file to load -- app_specific

gem dependency warbler
Gem warbler-2.0.0
  jbundler (~> 0.9, development)
  jruby-jars (>= 9.0.0.0)
  jruby-rack (< 1.3, >= 1.1.1)
  rake (>= 10.1.0)
  rdoc (>= 2.4.2, development)
  rspec (~> 2.10, development)
  rubyzip (< 1.4, ~> 1.0

It works fine with a earlier warbler version which targets jruby <=2 (jruby 1.7.26).

gem dependency warbler      
Gem warbler-1.4.9
  jbundler (~> 0.5.5, development)
  jruby-jars (< 2.0, >= 1.5.6)
  jruby-rack (< 1.3, >= 1.1.1)
  rake (>= 0.9.6)
  rdoc (>= 2.4.2, development)
  rspec (~> 2.10, development)
  ruby-maven (~> 3.1.1.0, development)
  rubyzip (< 1.2, >= 0.9)

grtjn:
Related to #579..

grtjn:
Thanks for the report. Installing warbler with `gem install warbler -v 1.4.9` works best indeed. This line can also be commented out. No need to run with jruby, it comes with warbler:

https://github.com/marklogic/roxy/blob/master/deploy/lib/server_config.rb#L147

Issue #579, opened 2016-03-06
Exception creating roxy jar


joecrean:
This was a bug I previously discussed with @grtjn and @paxtonhare  - here finally is the bug report...sorry it took so long

Testing on my local Mac environment. When I run ``ml jar`` i see the following exception:

```
~/Workspace/Thieme/ps-src-main/core-services/trunk -> ml jar

ERROR: You must be using JRuby to create a jar
Usage: ml jar

General options:
  -v, [--verbose]                   # Verbose output

Prerequisites:
  - You must be running JRuby http://jruby.org/
  - You must have the warbler gem installed
    > gem install warbler
```
This makes no sense to me - why should i be running ``JRuby`` to create the roxy jar file. The only pre-requisite should be that I have the ``jruby`` and ``warbler`` gems installed  - which i do...

```
~/Workspace/Thieme/ps-src-main/core-services/trunk -> gem list

*** LOCAL GEMS ***

bigdecimal (1.2.4)
bundler (1.6.2)
bundler-unload (1.0.2)
executable-hooks (1.3.2)
gem-wrappers (1.2.4)
io-console (0.4.2)
jruby-jars (9.0.1.0, 1.7.22)
jruby-rack (1.1.19)
json (1.8.1)
minitest (4.7.5)
psych (2.0.5)
rake (10.1.0)
rchardet (1.6.1)
rdoc (4.1.0)
rubygems-bundler (1.4.4)
rubygems-update (2.2.3)
rubyzip (1.1.7)
rvm (1.11.3.9)
test-unit (2.1.2.0)
warbler (1.4.9)
```

In order to workaround this problem i just commented out the first line in the ``self.jar`` method in ``server_config.rb`` - see below...

```
  def self.jar
    # raise HelpException.new("jar", "You must be using JRuby to create a jar") unless RUBY_PLATFORM == "java"
    begin
      # ensure warbler gem is installed
      gem 'warbler'
      require 'warbler'

      jar_file = ServerConfig.expand_path("#{@@path}/../roxy.jar")
      logger.debug(jar_file)
      Dir.mktmpdir do |tmp_dir|
        logger.debug(tmp_dir)

        temp_roxy_dir = tmp_dir + "/roxy"
        Dir.mkdir temp_roxy_dir

        FileUtils.mkdir_p temp_roxy_dir + "/bin"
[...]

PR #887, opened 2018-08-13
Issue 579 exception creating roxy jar

rlouapre:
Following changes are require to rebuild roxy.jar using warbler `2.0.5`
JRuby 9.x requires Java 8
Related #659


grtjn:
#579

grtjn:
This looks good. Is there a quick way to run a test with this? I've not generated a roxy jar for a while.

@RobertSzkutak wdyt?

Issue #862, opened 2017-11-23
Unable to deploy modules if XCC port is secured with TLS


mwarnes:
RFE

Roxy is unable to deploy modules if he XCC port is secure using TLS

Within  ./deploy/lib/xcc.rb the url is hardcoded to use http.

      def build_load_uri(target_uri, options, commit)
      url = "http://#{@hostname}:#{@port}/insert?"

It would be useful if the url was created with the protocol set to http our https depending on a configuration setting such as in ./deploy/lib/ml_rest.rb, i.e

     @rest_protocol = "http#{options[:use_https_for_rest] ? 's' : ''}"
     ...
     url = "#{@rest_protocol}://#{@hostname}:#{@port}/v1/config/properties"

The issue can be worked round by hardcoding https in the xcc.rb url.

grtjn:
I'm surprised this hasn't surfaced before yet. Shouldn't be too difficult to implement. Would you be willing to fork the project, and open a PR? You can find guidelines in [CONTRIBUTING.md](https://github.com/marklogic-community/roxy/blob/master/CONTRIBUTING.md)..

heelix:
Do you know if this one was merged in?  We are trying to add in a few SSL xcc ports to our app via ROXY (rather than tweak by hand).  I can pick up this ball if it did not go anywhere.

grtjn:
I don't think anyone looked into this yet, so feel free to open a PR!

heelix:
Will do.  Planning to get Roxy current in our environment as soon as 9.0-4.2 hits the downloads.  Will be testing on that.

PR #891, opened 2018-09-18
Switch the xcc port to https if https is enabled.

ciprianc:
When building the url to talk to the xcc port roxy doesn't look whether that should be https or not,
it's just hardcoded to http, which makes it impossible to deploy modules. This commit changes that
by switching to https if either `use-https-for-rest` is set to true or `ssl-certificate-template`
exists in the config.

First attempt at fixing #862

### DIFF
--- a/deploy/lib/server_config.rb
+++ b/deploy/lib/server_config.rb
@@ -2451,7 +2451,8 @@ private
           :http_connection_retry_count => @properties["ml.http.retry-count"].to_i,
           :http_connection_open_timeout => @properties["ml.http.open-timeout"].to_i,
           :http_connection_read_timeout => @properties["ml.http.read-timeout"].to_i,
-          :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i
+          :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i,
+          :use_https_for_xcc => @properties["ml.use-https-for-xcc"] == "true"
         })
       end
   end
--- a/deploy/lib/xcc.rb
+++ b/deploy/lib/xcc.rb
@@ -100,6 +100,7 @@ module Roxy
       })
       @request = {}
       @gmt_offset = Time.now.gmt_offset
+      @xcc_protocol = "http#{options[:use_https_for_xcc] ? 's' : ''}"
     end
 
     def xcc_query(options)
@@ -176,7 +177,7 @@ private
     end
 
     def build_load_uri(target_uri, options, commit)
-      url = "http://#{@hostname}:#{@port}/insert?"
+      url = "#{@xcc_protocol}://#{@hostname}:#{@port}/insert?"
 
       url << "uri=#{url_encode(target_uri)}"
 
--- a/deploy/default.properties
+++ b/deploy/default.properties
@@ -202,6 +202,11 @@ use-https=false
 #
 use-https-for-rest=false
 
+#
+# set this to true if you are using https for your xcc servers
+#
+use-https-for-xcc=false
+
 ## Security
 #
 # The authentication used for your appserver


deploy/lib/server_config.rb
```diff
@@ -2451,7 +2451,8 @@ def xcc
           :http_connection_retry_count => @properties["ml.http.retry-count"].to_i,
           :http_connection_open_timeout => @properties["ml.http.open-timeout"].to_i,
           :http_connection_read_timeout => @properties["ml.http.read-timeout"].to_i,
-          :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i
+          :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i,
+          :use_https_for_rest => @properties["ml.ssl-certificate-template"].present? || @properties["ml.use-https-for-rest"] == "true"
```
grtjn: The rest server that gets called by Roxy is typically the server at app-port, to which ssl-cert stuff is typically applied. The xcc server could be a separate server, and it is more difficult to guess if it is running https or not. Easiest way to solve that would be to introduce a `use-https-for-xcc` property in default.properties, which can be overridden in the other ones. The `use_https_for_rest` (or better: `use_https_for_xcc`) would only respond to that property in the case of XCC, unless perhaps the `install-xcc` property has been set to false, and xcc-port equals app-port (a helper function to verify this would be convenient).

deploy/lib/xcc.rb
```diff
@@ -100,6 +100,7 @@ def initialize(options)
       })
       @request = {}
       @gmt_offset = Time.now.gmt_offset
+      @rest_protocol = "http#{options[:use_https_for_rest] ? 's' : ''}"
```
grtjn: rest --> xcc

deploy/lib/xcc.rb
```diff
@@ -176,7 +177,7 @@ def go(url, verb, headers = {}, params = nil, body = nil)
     end
 
     def build_load_uri(target_uri, options, commit)
-      url = "http://#{@hostname}:#{@port}/insert?"
+      url = "#{@rest_protocol}://#{@hostname}:#{@port}/insert?"
```
grtjn: rest --> xcc

grtjn:
Unfortunately I can't really test with https enabled easily, but it looks like the changes are made in the right places. I left a few suggestions inline though..

deploy/lib/server_config.rb
ciprianc: I wasn't sure whether to use the `use-https-for-rest` or just the `use-https` option for xcc, so I went with what #862 mentioned. I agree with you though, it's best to introduce a separate option for xcc in particular. I'll amend the PR.

deploy/lib/server_config.rb
```diff
@@ -2451,7 +2451,8 @@ def xcc
           :http_connection_retry_count => @properties["ml.http.retry-count"].to_i,
           :http_connection_open_timeout => @properties["ml.http.open-timeout"].to_i,
           :http_connection_read_timeout => @properties["ml.http.read-timeout"].to_i,
-          :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i
+          :http_connection_retry_delay => @properties["ml.http.retry-delay"].to_i,
+          :use_https_for_xcc => @properties["ml.use-https-for-xcc"] == "true"
```
ciprianc, line 2455: So the similar `use-https-for-rest` option checks if either `use-https-for-rest` is `true` _OR_ `ssl-certificate-template` is defined. In that case it makes sense because as soon as you define a certificate template it turns the rest server to `https` when you bootstrap, you don't have to also enable `use-https-for-rest`. For XCC the xml for the definition is built a bit differently, and defining a certificate-template won't also enable ssl on the xcc port. If I were to have the same _OR_ condition I think it would break backwards compatibility. Let me know what you think.

grtjn: Yes, that was exactly the concern I had before. Looks perfect to me..

grtjn:
I'll merge, so that our internal CI can run the self-test with it.

